PR #2, opened 2019-10-02
Implementing dark mode

bastienwirtz:
Dark mode theme based on OS preference, with support of user override.


twolaw:
Could optional message background be darkened the same way ?
And I suggest `body #app.is-dark .subtitle { color: #b5b5b5; }` for greyed subtitles.

And could bighead backgroud colors be chosen in config.yml ?

Issue #9, opened 2019-12-20
Service worker does not properly cache fonts


fbartels:
Hi @bastienwirtz,

I have been playing around with your start page a bit and in my test environment I am actually running it in a subdir (I have opened https://github.com/bastienwirtz/homer/pull/8 to fix some code related to that).

Once thing I am unable to fix myself is the following:

When switching in the dev console to offline mode, all externally added fonts are missing. Which means for example the font awesome icons, but also the title looks different (which is a google font as far as I can see). I have found https://stackoverflow.com/questions/39432717/how-can-i-cache-external-urls-using-service-worker which sounds like it could help to improve, but I am not sufficient enough with Javascript to actually implement the solution.

bastienwirtz:
Hi Felix, 
Thanks for reporting that, I'll look into it very soon :+1:

fbartels:
Hi Bastien,

thanks. In of of the forks of the repo has added these resources locally, which means it could be easier to add them to the manifest.

bastienwirtz:
Hi @fbartels!
This should do the trick: https://github.com/bastienwirtz/homer/pull/11

fbartels:
Almost. I have commented the pull request.

Thanks so far!

PR #11, opened 2020-01-13
Adding full off network support.

bastienwirtz:
Related to https://github.com/bastienwirtz/homer/issues/9. 
Copy all external resources locally to make it usable full offline.


fbartels:
Hi, thanks. Looks quite a bit better now. Sometimes a picture amounts for a thousand words:

Screenshot 2020-01-13 at 08 22 25

When switching Chrome to offline after a first normal request then it almost looks good, only the image above the "You're offline bro" is missing.

Screenshot 2020-01-13 at 08 22 45

When now doing a forced reload however (ctrl+shift+r) some more resources go missing. Not quite sure if this can be labeled as expected behaviour, but then it would still be strange that other resources survive this treatment.

bastienwirtz:
My bad, I forgot the additional font awesome files in the service worker, it should be better now!

fbartels:
Thanks again. Now the only resources throwing errors are the `?alive` request (which I would label as expected) and the `favicon.png` request.

bastienwirtz:
Good catch, the favicon was missing too. Thanks to you to contribute and taking some time to test it!
Have a nice day !

PR #12, opened 2020-01-21
Dockerfile

lukasmrtvy:
Interested in this simple Dockerfile ?
Would be also great to setup official docker hub repository ( https://docs.docker.com/docker-hub/builds/ ).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,14 @@ FROM alpine:3.11
 
 COPY ./ /www/
 
-ENV USER abc
-ENV GROUP abc
+ENV USER darkhttpd
+ENV GROUP darkhttpd
 ENV GID 911
 ENV UID 911
 
 RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} && \
     apk add -U darkhttpd
 
-USER abc
+USER darkhttpd
 
 ENTRYPOINT ["darkhttpd","/www/"]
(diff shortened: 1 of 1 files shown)


fbartels:
You could make the resulting image even smaller by using https://github.com/halverneus/static-file-server as the base image.

fhemberger:
@fbartels Dockerfile would be nice and short this way:

```dockerfile
FROM halverneus/static-file-server
COPY . /web
```

Web server is exposed by default on port 8080.

bastienwirtz:
Hi,
Thanks all of you! I'll look into it & test it asap. 
@fbartels @fhemberger the static file server image looks nice but the initial proposition is still quite small and clean, and as it come straight from the official base alpine image, it removes a level of dependency, it's a solid choice for maintainability.

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:3.11
+
+COPY ./ /www/
+
+ENV USER abc
```
fhemberger: Nit: Could you please rename the user/group to `darkhttpd`?

bastienwirtz: +1 That would be great!

lukasmrtvy:
I can add s6 overlay for runtime uid,gid change , if You want to implement some crud to local config in the future to support proper volume handling..

Dockerfile
lukasmrtvy: done

bastienwirtz:
Finally got to test it, looks good :+1: 
Thanks!

calvinbui:
@bastienwirtz could you enable the Docker Hub builds?

PR #16, opened 2020-02-09
Display services notification badges directly on cards

twolaw:
Using their API, it can fetch and display notification badges of those services :  radarr, sonarr and medusa.
Maybe more services could be added.

Example:
image

### DIFF
--- a/app.css
+++ b/app.css
@@ -188,7 +188,7 @@ body {
     color: #4285f4;
     background-color: #4285f4;
     position: absolute;
-    top: 1rem;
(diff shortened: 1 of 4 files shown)


twolaw:
No problem, I understand.
I made it for me first and thought maybe others could be interested.

Btw, it was the first time I write javascript code, must be far from clean. 😊

bastienwirtz:
Thanks for understanding @twolaw, and I keep this idea in mind!

PR #22, opened 2020-03-12
Update <head/><title/> to 'title' defined in config.yml

traeblain:
Added some simple javascript to update the page <title></title> to the same as the configuration.

**Note:** Did this after the getConfig(), but due to ServiceWorker and offline/online changes, I wasn't sure if a watcher should have been used to monitor.  Unfortunately, I'm not all that aware of how things work when offline.  In case it's necessary, there the code that could be used:

```js
({
methods: { 
...
},
watch: {
    config: function (val, old) {
        document.title = val.title + ' | Homer';
    }
}
});
````

### DIFF
--- a/app.js
+++ b/app.js
@@ -20,6 +20,7 @@ const app = new Vue({
         this.checkOffline();
         try {
             this.config =  await this.getConfig();
+            document.title = this.config.title + ' | Homer';
         } catch (error) {
             this.offline = true;
         }


bwirtz:
Wow @traeblain, just seen all your contributions, big, big thanks!

This looks good to me! I might change the way meta data are managed in the near future to be able to add a webmanifest to make it a PWA (among other thing, you could "install" / add to your home screen the page, which could be nice). The manifest will needs to include the title, colors... and no code can be use in it. So I will have to add a command to generate it (haven't thought a lot about it yet but I'll try to keep it light and simple, it's the philosophy of the project, maybe avoid to setup a full build system with something like webpack)

bastienwirtz:
Wow @traeblain, just seen all your contributions, big, big thanks!

This looks good to me! I might change the way meta data are managed in the near future to be able to add a webmanifest to make it a PWA (among other thing, you could "install" / add to your home screen the page, which could be nice). The manifest will needs to include the title, colors... and no code can be use in it. So I will have to add a command to generate it (haven't thought a lot about it yet but I'll try to keep it light and simple, it's the philosophy of the project, maybe avoid to setup a full build system with something like webpack)

Issue #21, opened 2020-03-04
[FEATURE] Removal of the Top "Blue Bar" and option to retain the search, theme and view switch options without having the Page link option


mokshmridul:
Hi,

I wanted to make a few feature requests, namely:

1) Option to remove the following section from the page, and consequentially remove the entire blue bar which appears on the top part of the page

```
title: "Moksh's Homelab"
subtitle: "Homelab"
logo: "assets/logo.png"
 icon: "fas fa-skull-crossbones" Optional icon
footer: false
```

2) I would also like the option to retain the search, view switch and theme switcher to remain, even if no pages are referenced in the following section:

```links:
  - name: "Home Page"
    icon: "fab fa-github"
    url: "https://mokshmridul.in"
    target: '_blank' # optional html a tag target attribute
```

Thanks a ton

DaPutzy:
You can do this:

    links: {}

to keep the bar without links in it.

bastienwirtz:
Hi @mokshmridul! Thanks for the feedback!
The header hiding is on it's way here: https://github.com/bastienwirtz/homer/pull/26 :+1:

mokshmridul:
> Hi @mokshmridul! Thanks for the feedback!
> The header hiding is on it's way here: #26 👍

WOW !!

Thanks a ton mate, looking forward to the merge :)

bastienwirtz:
@mokshmridul took a bit more time than expected and have been moved in another pull request, but it's still coming: https://github.com/bastienwirtz/homer/pull/26! Stay tuned!

mokshmridul:
> @mokshmridul took a bit more time than expected and have been moved in another pull request, but it's still coming: #26! Stay tuned!

Thank you so much mate

Looking forward to this change :)

bastienwirtz:
Hello @mokshmridul, It has been merge, hope that works for you!

mokshmridul:
> Hello @mokshmridul, It has been merge, hope that works for you!

Hi, Thank you so much for this. Ill check and close this thread once the docker file gets udpated.

Thank you once again.

PR #25, opened 2020-03-12
Documents the hidden feature to maintain navbar w/o links.

traeblain:
Simply adds a line to `config.yml` and `README.md` to show how to keep the navbar but without any links.

Addresses the second part of #21 

Used `[]` over `{}` because that is what is expected from the YAML conversion.


bastienwirtz:
Perfect! Thanks!!

Issue #13, opened 2020-01-22
Make navbar responsive (toggle into hamburger menu on mobile)


twolaw:
Instead of hiding on mobile, it would be interesting that navbar-menu toggles into navbar-burger.

Bulma documentation : https://versions.bulma.io/0.7.0/documentation/components/navbar/

bastienwirtz:
Hi @twolaw!
Thanks for opening an issue. It would be nice to have a narbar-burger. I've some stuff I want to do before but I keep that in my todo list!

bastienwirtz:
Hi there! just merged @traeblain's PR that bring the navbar-burger, enjoy!

PR #27, opened 2020-03-14
Added Hamburger Menu for small screens per Bulma spec.

traeblain:
Think I got this right.

Added the details per the Bulma spec.  Had to fiddle a bit as I believe (since they're icons only) the dark mode, layout, and search are expected to be on the same line even in the phone layout.  This takes care of it and gets back to pretty close to what you had originally.

Addresses Issue #13

### DIFF
--- a/app.css
+++ b/app.css
@@ -1,277 +1,372 @@
+@charset "UTF-8";
 /* raleway-regular - latin */
 @font-face {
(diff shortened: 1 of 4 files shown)


bastienwirtz:
Awesome! thanks!

PR #37, opened 2020-03-26
Add keyboard shortcuts to navigate to the first search result.

jozefs:
## Description

Added the following keyboard shortcuts:

* `Enter` Open the first matching result in the active tab.
* `Alt`/`Option` + `Enter` Open the first matching result in a new tab.

In some browsers (Firefox) opening the result in a new tab might be blocked initially.

The filtering has been deduplicated and extracted into its own method. I have not been able to figure out how to get the first result without iterating over the items again but the iteration stops as soon as the first result is found.

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,13 @@ A dead simple static **HOM**epage for your serv**ER** to keep your services on h
 
 **Check out the live demo [here](https://homer-demo.netlify.com/).**
 
+It supports keyboard shortcuts:
+
+* `/` Start searching.
+* `Escape` Stop searching.
+* `Enter` Open the first matching result (respects the bookmark's `_target` property).
+* `Alt`/`Option` + `Enter` Open the first matching result in a new tab.
+
 If you need authentication support, you're on your own (it can be secured using a web server auth module or exposing it only through a VPN network / SSH tunnel, ...)
 
 ![screenshot](https://raw.github.com/bastienwirtz/homer/master/screenshot.png)
--- a/app.js
+++ b/app.js
@@ -84,6 +84,26 @@ const app = new Vue({
         },
         toggleMenu: function() {
             this.showMenu = !this.showMenu;
+        },
+        matchesFilter: function(item) {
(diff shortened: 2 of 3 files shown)


traeblain:
Just a thought, but shouldn't `Enter` honor the `target=""` definition in the config?  If the expectation is to open in a new window from the config, shouldn't this also follow the expectation?

Wibbbs:
> 
> 
> Just a thought, but shouldn't `Enter` honor the `target=""` definition in the config? If the expectation is to open in a new window from the config, shouldn't this also follow the expectation?

Agreed

jozefs:
That is an interesting point. In my use case, whether I want to open a bookmark in the current tab or a new tab depends on context. For example if I want to open a single bookmark I always open it in the same tab. But if I’d like to open multiple bookmarks I typically open them in new tabs, even if one of them is the same bookmark from the previous scenario.

traeblain:
I think the multiple opinions is why I suggest that `Enter` honor the config. 

The best (or compromise...or worst...) plan here might be that `Enter` follows the config, and `Alt/Option+Enter` opens in a new tab/window regardless of the config.  Thought is this: in your use case, the assumption is that you don't have any `target="_blank"` definitions in your config because the need for a new tab is context based.  Others will have a `target="_blank"` defined because they have a different use case and want control over when the new window opens and when it doesn't.  That way it sort of envelopes the intents and operates in the exact same manner a `click` or `ctrl-click` would operate.

jozefs:
OK. Let's wait for any feedback from @bastienwirtz and I'll send an update to this pull request incorporating all requested changes.

bastienwirtz:
Hello, 

> Enter follows the config, and Alt/Option+Enter opens in a new tab/window regardless of the config.

:+1:  I like the idea it operates in the same way the browser do, let's do that! 

Thanks all of you!

jozefs:
Updated the pull request. `Enter` now respects the bookmark's `target` property.

PR #38, opened 2020-03-26
Added "GoLinks" functionality

beyertom:
### What is "GoLinks"?
You can set Homer as the default search engine in your browser.
For example, if you search for the search term "g how high is the Leaning Tower of Pisa", this search will be performed at Google.
If you replace the "g" with a "d", the search will be performed at DuckDuckGo.

I think this change makes Homer much more versatile

## Type of change

## Checklist:


AlexanderEDavis:
Do you think you'd be able to set it so you dont have to add g or d. Could that be a setting in the config?
Also, is there a way to let it search both your homer links and the web at the same time?

klejejs:
Any progress on this?

bastienwirtz:
HI there, sorry for letting this PR stale. Thanks for your contribution @beyertom! 
This PR adds lots of configuration line in the settings file and the goal of Homer is to be simple and focus on a single purpose which is expose internal service (Golinks is not required for that). That being said, it could be a nice addition, but the code base has changed a lot, I'm gonna close this one and I will think about a way to integrate it in the future.

PR #52, opened 2020-05-16
Slim down Dockerfile

fbartels:
## Description

This PR slims down the Dockerfile by using https://github.com/halverneus/static-file-server instead of Alpine with an added webserver. static-file-server is built in go and can run in a scratch image.

To be backwards compatible the Dockerfile uses the same directory to serve assets from as the previous container.

Additionally this includes a docker-compose.yml to simplify local building and running.

## Type of change

- [*] New feature (non-breaking change which adds functionality)

## Checklist:

- [*] I read & comply with the [contributing guidelines](https://github.com/bastienwirtz/homer/blob/master/CONTRIBUTING.md)
- [*] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. 
- [*] I've check my modifications for any breaking change, especially in the `config.yml` file

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@ dockerfile
 *.md
 .git
 screenshot.png
+docker-compose.override.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,8 @@
-FROM alpine:3.11
+FROM halverneus/static-file-server:v1.8.0
+
+ENV \
+	DEBUG=true \
+	FOLDER=/www \
+	PORT=8080
 
 COPY ./ /www/
-
-ENV USER darkhttpd
-ENV GROUP darkhttpd
-ENV GID 911
-ENV UID 911
-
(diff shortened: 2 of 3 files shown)


DocBrown101:
What kind of problem will this solve? All I see is that another dependency is added!

fbartels:
@DocBrown101 that made me think a bit. The resulting image is actually slightly bigger, 11.86MB compared to 10.58MB for the current one.

I would not say that this adds a dependency in general. strictly speaking it replaces the current Alpine and darkhttp dependency with a single dependency on static-file-server.

DocBrown101:
Almost every Docker user knows Alpine. I hear about static-file-server for the first time.

In general I think that the file size is already very slim and therefore no adjustment is justified.

bastienwirtz:
Hi there! 
I appreciate the intent @fbartels! But yes Alpine is really slim already, and most important it's the docker official recommendation and well maintained, that's safe and efficient, unless there is a huge advantage to change, I'd rather keep Alpine.

fbartels:
Sure, no problem. Let me know if I should open a new pr only for the docker-compose.yml.

PR #49, opened 2020-05-08
Build multi-arch container with manifest

nightah:
## Description

This change will publish containers for the following Architectures:
- amd64 (x86_64)
- arm32v7 (armhf)
- arm64v8 (aarch64)

The `latest` manifest will be published on DockerHub which when pulled by a user, will distribute the respective container for the appropriate architecture.

This will require changes in the auto-build section of DockerHub per the below image:

dockerhub

Build caching can be turned off.

Fixes #40 and fixes #43.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] I read & comply with the [contributing guidelines](https://github.com/bastienwirtz/homer/blob/master/CONTRIBUTING.md)
- [ ] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. 
- [ ] I have made corresponding changes the documentation (README.md).
- [ ] I've check my modifications for any breaking change, especially in the `config.yml` file

PR #56, opened 2020-05-20
Fix docker multiarch builds

nightah:
This is a continuation/fix for #49.

### DIFF
--- a/Dockerfile.arm32v7
+++ b/Dockerfile.arm32v7
@@ -1,4 +1,5 @@
 FROM arm32v7/alpine:3.11
+COPY qemu-arm-static /usr/bin
 
 COPY ./ /www/
 
--- a/Dockerfile.arm32v7
+++ b/Dockerfile.arm32v7
@@ -1,6 +1,12 @@
-FROM arm32v7/alpine:3.11
-COPY qemu-arm-static /usr/bin
+FROM alpine as qemu
 
+ARG QEMU_VERSION "v4.2.0-7"
+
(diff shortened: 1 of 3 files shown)


Dockerfile.arm32v7
```diff
@@ -1,12 +1,13 @@
-FROM alpine AS builder
+FROM alpine as qemu
 
-# Download QEMU, see https://github.com/docker/hub-feedback/issues/1261
-ENV QEMU_URL https://github.com/balena-io/qemu/releases/download/v3.0.0%2Bresin/qemu-3.0.0+resin-aarch64.tar.gz
-RUN apk add curl && curl -L ${QEMU_URL} | tar zxvf - -C . --strip-components 1
+ARG QEMU_VERSION "v4.2.0-7"
```
bastienwirtz, line 3: @nightah  `ENV QEMU_VERSION "v4.2.0-7"` ?

bastienwirtz: Got a: 
```
Unexpected error
Encountered error: 400 Client Error: Bad Request ("Dockerfile parse error line 3: ARG requires exactly one argument")
```

nightah: No an ARG is fine it doesn't need to persist beyond the download.

nightah: https://vsupalov.com/docker-arg-vs-env/ just FYI.

bastienwirtz: So might be `ARG QEMU_VERSION="v4.2.0-7" to solve the parse error ? (https://docs.docker.com/engine/reference/builder/#default-values)

nightah: Yep the above is correct syntax, if you can correct it that'd be great thanks.

Issue #58, opened 2020-05-21
Tab Bar Title


AlexanderEDavis:
Is there a way to remove or customise the Homer title in the tab bar?

![image](https://user-images.githubusercontent.com/9097891/82600238-84508500-9ba5-11ea-89a7-8c5f201b110e.png)

fbartels:
There is a fork that implements a feature to configure this, not quite sure why the author of it did not (yet?) make a pr with it. I have pulled it into my local copy a few days ago and it seems to work without problems.

https://github.com/sylv-io/homer/commit/1ba17a9b2c2fef58a566e4850d40620f03e9f0e8

CC @sylv-io

sylv-io:
oh sure i can do it :+1:

bastienwirtz:
👋
Good idea! Thanks!

PR #60, opened 2020-05-22
app.js: html title based on configuration

sylv-io:
Use the title and subtitle defined in the configuration file as
html title.

Signed-off-by: sylv <sylv@sylv.io>

## Description
This PR refers to #58. With this change the html title is a combination of title + subtitle, instead of
title + hardcoded 'Homer". 

## Type of change

## Checklist:

### DIFF
--- a/app.js
+++ b/app.js
@@ -21,7 +21,7 @@ const app = new Vue({
         this.checkOffline();
         try {
             this.config =  await this.getConfig();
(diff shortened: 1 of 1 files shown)


carlox97:
If we disable the subtitle we still have "Title | ". Wouldn't it be better to have just "Title" if we don't use a subtitle?

Issue #86, opened 2020-06-12
This docker repo is a mess


BodingClockchian:
Great work keeping this up but its a total mess.
Manually copying the config files from the localmachine? The dockerfile should be configured to do this.

bastienwirtz:
Hi @BodingClockchian, 
I would not qualify this particular issue a "total mess" (unless I miss something else?). But anyway, that's planned, I will fix that after https://github.com/bastienwirtz/homer/pull/87 is merged.

GlennToms:
@BodingClockchian Hopefully my merge has made Homer with docker easier for you to use.  Could you close this issue if your happy? Many thanks.

PR #87, opened 2020-06-12
Added ENV PORT options and EXPORT PORT

GlennToms:
## Description

Allows user to set the port they want to run the web server on when using docker.

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,12 +21,10 @@ ENV PORT 8080
 RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} && \
     apk add -U darkhttpd
 
-RUN echo "darkhttpd /www/ --no-listing --port $PORT" > /entrypoint.sh
-RUN set -ex chown ${USER}:${GROUP} /entrypoint.sh
+COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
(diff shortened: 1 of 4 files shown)


bastienwirtz:
Hi @GlennToms!
Thanks for your contribution! It looks good to me, but I think it would be even better to create the `entrypoint.sh` in the repo, it might be usefull. I think I will add the assets initialization to the entrypoint (https://github.com/bastienwirtz/homer#using-docker, the `cp` part) if not exists. 
How does that sounds ? 

Might be great to regroup all docker related file in a docker directory, but not sure if we can build the container if the Dockerfile is in the root directory? I'll look into it.

GlennToms:
Thanks @bastienwirtz
The cp command in Alpine doesn't have the '-n' arg. So the below will copy files that don't exist.

entrypoint.sh

```
#!/bin/sh

yes n | cp -i /www/config.yml.dist /www/config.yml
while true; do echo n; done | cp -Ri /www/assets /www/assets 2>/dev/null

darkhttpd /www/ --no-listing --port $PORT
```

GlennToms:
@bastienwirtz 
Branch is now ready.  entrypoint script created and dockerfile(s) have been updated.
I've tested it and its working as intended.
#86 Homer now launches the default/sample [config.yml](https://github.com/bastienwirtz/homer/blob/master/public/config.yml.dist) without any user configuration needed.

PR #89, opened 2020-06-13
Added ability to hide the message

Davy1992:
## Description

The current codebase does not hide the message block, even when the title and content is empty the message although invisible is still taking up space at the top of the page. So I've added the 'hide' class that hides the message block allowing you to control the visibility of it via the endpoint.

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ colors:
 # Optional message
 message:
   # url: "https://<my-api-endpoint>" # Can fetch information from an endpoint to override value below.
-  style: "is-warning"
+  style: "is-warning" # 'hide' can be added to hide the message block until endpoint overrides it.
   title: "Optional message!"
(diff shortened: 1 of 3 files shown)


bastienwirtz:
Hi @Davy1992 !

Thanks for your contribution!
I don't think we need this, because the message can be hidden just by commenting the configuration file, or setting null or blank value for title & message. **Or it's supposed to be! Thanks to your PR I've found a bug** in the message component initialization when using no default values from the config file, just pushed a fix: https://github.com/bastienwirtz/homer/commit/7ef65940eef485d77ae5e7eb4bb6fbb1e431dd71

It should works fine now, let me know if you have any issues.

By the way, visibility control class already exists, available via Bulma (the UI kit used in this project): https://bulma.io/documentation/helpers/visibility-helpers/.

Davy1992:
No issues here. Thanks for the visibility helpers they are what I needed for my use case.

Issue #83, opened 2020-06-09
Docker installation, config.yml being bound to directory?


blasphemite:
Using the docker run command from the readme, changing the directories and adding the `-d` option.

`docker run -d -p 8803:8080 -v /container/config.yml:/www/config.yml -v /container/assets/:/www/assets b4bz/homer:latest`

(copy assets directory to local assets directory)

Local /container/config.yml is created as a directory. 

If I remove the local config.yml directory and manually replace it with a default config file, I get errors about binding a volume to a file.

If I create the config.yml prior to creating the container, I get a "404 page not found" error.

JamiePhonic:
Try this:

first, create a `homer` folder inside your `/container/` folder and place the files inside it:
```
/container/homer/assets/
/container/homer/config.yml
```
then run:
```sh
chmod -R 774 /container/homer/
docker run -d --name='homer' -p '8066:8080' -v '/container/homer/assets/':'/www/assets':'rw' -v '/container/homer/config.yml':'/www/config.yml':'rw' 'b4bz/homer:latest' 
```

blasphemite:
I actually do have the homer files in it's own folder. The directory is more like `/docker/homer/assets/` etc.

chown and chmod on homer folder, then running:

`docker run -d --name='homer' -p '8066:8080' -v '/docker/homer/assets/':'/www/assets':'rw' -v '/docker/homer/config.yml':'/www/config.yml':'rw' 'b4bz/homer:latest'`

and config.yml is still created as a directory.

Aimsucks:
Yes, he's saying you need to create `config.yml` inside `/docker/homer/` because volumes are meant to be mapped to folders and not files. It will work, however, if you try to mount an existing file to the container.

JamiePhonic:
Yes, if mapping a file, the file MUST exist before starting the container, otherwise docker will usually assume It's a folder.
Also make sure the file has the correct permissions so docker can access it.

Your're not the only one to experience this issue, for example, in #50, the OP was having the same issue running on a synology NAS. Creating the config.yml file before starting the container seemed to work.

On the other hand, I'm running on Unraid and never ran In to this issue at all.

 I think part of it may be down to how docker Is implemented/managed within a perticular OS, but thats just a theory.

bastienwirtz:
Hi there!

[...]

PR #94, opened 2020-06-24
Regroup all editable files in one place

bastienwirtz:
## Description

Regroup all editable files into one place, the `assets` directory. This makes things easier for docker installation, only one directory have to be binded. This change is retro compatible.

what's done: 
- Moved the `config.yml` to `assets`
- Fixed default assets copy

Fixes https://github.com/bastienwirtz/homer/issues/83


bastienwirtz:
Hey @GlennToms! 
I made a few adjustments on your recent changes, it would be awesome if you can have an have a look!

PS: Docker hub auto build are temporarily disabled, your previews changes to the docker files are not released on docker hub yet.

entrypoint.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/sh
 
-yes n | cp -i /www/config.yml.dist /www/config.yml
-while true; do echo n; done | cp -Ri /app/dist/www/assets /www/assets 2>/dev/null
+while true; do echo n; done | cp -Ri /www/default-assets/* /www/assets/
+
```
GlennToms, line 4: We need a way to launch a one of the default configs
You might want to swap out for **config.yml.dist.sample-sui** depending on what you want as the default page when a user starts on the container for the first time.

```suggestion
# This makes sure that at least the default config is available.
yes n | cp -i /www/assets/config.yml.dist /www/assets/config.yml
```

Dockerfile.arm64v8
```diff
@@ -33,8 +33,9 @@ RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} &
 
 COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
 COPY --chown=${USER}:${GROUP} entrypoint.sh /entrypoint.sh
+RUN mv /www/assets /www/default-assets
```
GlennToms, line 36:
```suggestion
Replace with `COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist/assets /www/default-assets`

File: Dockerfile.arm32v7 [hunk 3]
```diff
@@ -33,8 +33,9 @@ RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} &
 
 COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
 COPY --chown=${USER}:${GROUP} entrypoint.sh /entrypoint.sh
+RUN mv /www/assets /www/default-assets
```
GlennToms, line 36:
```suggestion
Replace with `COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist/assets /www/default-assets`

File: Dockerfile [hunk 4]
```diff
@@ -23,8 +23,9 @@ RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} &
 
 COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
 COPY --chown=${USER}:${GROUP} entrypoint.sh /entrypoint.sh
+RUN mv /www/assets /www/default-assets
```
GlennToms, line 26: We can't assume that **/www/assets** will be accessible to the UID that is running the build. On my system I get **access denied**

```suggestion
Replace with `COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist/assets /www/default-assets`

PR #96, opened 2020-06-25
Docker tweaks

gabe565:
## Description

No dependencies change with this PR. This just includes some small Docker changes to closer follow best practices.

- Adds `node_modules` to `.dockerignore`.
- Disables apk's cache while installing darkhttpd to shave ~1.5MB off of the final image.
- Removes `/www/config.yml` from the `VOLUME`s so that an empty directory isn't created when volume is not bound.
- Adds a workdir so that Docker always starts in `/www`
- Adds log output when the default config is copied over
- `exec` darkhttpd so that signals are passed to it

## Type of change

## Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@ dockerfile
 *.md
 .git
 screenshot.png
+node_modules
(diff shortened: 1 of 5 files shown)


bastienwirtz:
Hi @gabe565, 

Thanks for your contribution! 
I'm afraid there was already a PR open (that has been merge today) which include more or less the same modification and a few other tweaks:  https://github.com/bastienwirtz/homer/pull/94. I'm sorry for the duplicated work :(

gabe565:
Oops I didn't realize, you're right it looks like that PR made some of the exact same changes I had made! It's no problem, I'll rebase this and only leave a couple small changes that could still be beneficial. Thanks for the update!

PR #113, opened 2020-07-22
Add support for a custom stylesheet

gabe565:
## Description

This adds support for a custom stylesheet in config.yml.

I have been using it for importing Font Awesome Pro's CSS, but it could be used for individual styles too as an easier way than creating a custom theme.

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -27,8 +27,9 @@ connectivityCheck: true # whether you want to display a message when the apps ar
 theme: default # 'default' or one of the theme available in 'src/assets/themes'.
 
 # Optional custom stylesheet
-# Will load a custom CSS file. Especially useful for custom icon sets.
-# stylesheet: "assets/custom.css"
+# Will load custom CSS files. Especially useful for custom icon sets.
+# stylesheet:
+#   - "assets/custom.css"
 
 # Here is the exaustive list of customization parameters
 # However all value are optional and will fallback to default if not set.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -26,6 +26,10 @@ connectivityCheck: true # whether you want to display a message when the apps ar
 # Optional theming
 theme: default # 'default' or one of the theme available in 'src/assets/themes'.
 
+# Optional custom stylesheet
+# Will load a custom CSS file. Especially useful for custom icon sets.
+# stylesheet: "assets/custom.css"
+
 # Here is the exaustive list of customization parameters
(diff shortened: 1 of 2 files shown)


mistic100:
Very interested in the feature, allows to customize the app and still use the base docker image.

gabe565:
> Very interested in the feature, allows to customize the app and still use the base docker image.

Exactly. I've been using it to add FontAwesome Pro without having to do any rebuilding.

It isn't perfect since the stylesheet gets applied right after the initial page render so there will be a split second where the custom styles aren't applied. But with the right caching, it hasn't too big of a deal for me.

docs/configuration.md
```diff
@@ -26,6 +26,11 @@ connectivityCheck: true # whether you want to display a message when the apps ar
 # Optional theming
 theme: default # 'default' or one of the theme available in 'src/assets/themes'.
 
+# Optional custom stylesheet
+# Allows for @import statements and any other CSS. Useful for custom icon sets.
+# stylesheet: >-
+#   @import "assets/css/all.min.css";
```
bastienwirtz, line 32: What about just: 
```yaml
stylesheet: "assets/css/customization.css"
```
\+ on line 164: 

```js
this.createStylesheet(`@import  ${this.config.stylesheet}`);
```

To avoid mixing concept in `yaml` file and keep it pure configuration. Somebody less familiar with css and yaml could be confused seeing the `@import` here.

gabe565: I agree. I was originally thinking that people could add simple styles here which would load faster than a separate css file, but I guess the `colors` config covers most of that anyways. Thanks for the feedback!

gabe565:
Thank you! I pushed up your requested changes. Should be all around simpler now 👍

gabe565:
Actually, hold off on merging. I'm going to add support for the stylesheet config to be either a string or an array. That would be useful even if someone has multiple custom fonts and icons

bastienwirtz:
@gabe565 the only thing is, it's not cached by the service worker, so custom style won't apply offline, but I'll look at it later, we might be able to dynamically add it to the worker cache. Anyway, that's not a big deal.

gabe565:
Oh right. I've configured workbox to cache anything matching regex at runtime before. It might make sense to have the service worker cache everything in `assets/` here. Then custom icons would get cached offline too. The biggest downside is that by default, cache only gets busted when a build hash changes. We would need some way to manually trigger a refresh here.

PR #119, opened 2020-08-11
Feature/custom card colors

tpansino:
## Description

This PR adds basic support for customizing the background and text color of individual cards, via the `config.yml`.

This could also probably be implemented by making a custom style sheet, and allowing the `config.yml` to specify CSS classes to apply to each card, but I wanted to keep it simple. Happy to change my implementation if this is preferred.

I also ~didn't update the example configuration, but I'm happy to do that as well if desired.~ Edit: updated the `configuration.md`.

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -106,6 +106,9 @@ services:
         tag: "other"
         url: "https://www.reddit.com/r/selfhosted/"
         target: "_blank" # optionnal html a tag target attribute
+        # colors: # optional custom colors for card
+        #   background: #66a5e2
+        #   text: #000000
 ```
 
 If you choose to fetch message information from an endpoint, the output format should be:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -110,10 +110,8 @@ services:
         subtitle: "Another example"
         tag: "other"
         url: "https://www.reddit.com/r/selfhosted/"
-        target: "_blank" # optionnal html a tag target attribute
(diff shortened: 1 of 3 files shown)


bastienwirtz:
Hi @tpansino !
Thanks for your contribution! I like the idea. I just merged the support for custom CSS file by @gabe565 (https://github.com/bastienwirtz/homer/pulls?q=is%3Apr+is%3Aclosed) so I think the CSS class would be better, and provide more flexibility. 
If you are still willing to change the implementation for the CSS class system, go for it! that would be perfect :)

tpansino:
Agreed @bastienwirtz .

I have updated the feature to use custom CSS classes and created a `custom.css.sample` in the `public/` as an example of how to use it. Tested on Firefox and Chromium and it seems to look ok.

I am very much a backend developer :sweat_smile:, so please feel free to make suggestions if my CSS can be improved.

bastienwirtz (approved):
Well done @tpansino, it looks good!

bastienwirtz:
@tpansino On second look, I simplified a bit the service component: I moved your class to the parent element and removed it from the children. You can target any element from the parent class (example: `body #app .card.green title`) so it's ok to put it only once on the parent élément. You can find the diff here: https://github.com/bastienwirtz/homer/commit/ffe3404a2a9b36d69e880baa224375e82e64ae54. 

Does that still works for you ?

Issue #126, opened 2020-09-02
Maskable icon and iOS icon are too big and transparent


NotWoods:
The current maskable icon is just a copy of the normal icon. Maskable icons need the important content to be within the inner 80% of the image, and should be opaque rather than transparent.

Currently the icon [looks like this on a home screen](https://maskable.app?demo=https://raw.githubusercontent.com/bastienwirtz/homer/master/public/assets/icons/android-chrome-maskable-512x512.png)

I think both the maskable icon and iOS icon would look better if the gradient was applied to the entire background, not just the donut outline.

bastienwirtz:
Hey @NotWoods does this one is resolved with https://github.com/bastienwirtz/homer/pull/127 ?

NotWoods:
Yup!

PR #127, opened 2020-09-03
Reduce number of PWA icons, fix paths

### LINKED ISSUE #124: [Bug] PWA not working in Firefox
I'm trying to save the app as a PWA on my Pixel 3A XL, running Android 10 August security patch, using Firefox 68.11.0.

The PWA save icon shows up at the top of the browser as expected. Tapping it shows the save confirmation pop-up, however the Homer icon is missing. Tapping Save then does nothing except close the pop-up.

I'm happy to provide screenshots on request, however I suspect this will be easy to reproduce.

Edit: I'm not a frontend developer, but I have encountered issues with PWAs like this before, and it had something to do with the Vue configuration. Maybe we start by checking that?

NotWoods:
## Description

Rather than supplying a bunch of icon variations, its a lot easier to just have 512x512 icons and let other platforms resize the icon.

Maskable icons and iOS icons are now set up properly with correct sizing and a background. [Preview here](https://maskable.app/?demo=https://raw.githubusercontent.com/NotWoods/homer/pwa-icons/public/assets/icons/icon-maskable.png).

Also added a monochrome icon by reusing the Safari Pinned Tab icon.

Paths are resolved relative to the manifest URL.

Fixes #124, #126

## Type of change

## Checklist:


NotWoods:
Hi @bastienwirtz, please take a look when you have a chance. Happy to answer any questions and make changes.

bastienwirtz:
@NotWoods Looking at it right now! Thanks for your contribution! 
I'm gonna do some test, just to make sure the path work either in docker or direct. 

I know there is a lot of icons, but those can be auto generated as described here: https://github.com/bastienwirtz/homer/blob/master/docs/configuration.md#pwa-icons. 
On the other hand, it's kinda useless to have all individual size, I like your approach, it's more simple :thinking:

vue.config.js
```diff
@@ -1,103 +1,65 @@
...
-      .loader("raw-loader")
+      .use('raw-loader')
+      .loader('raw-loader')
       .end();
   },
-  publicPath: "",
+  publicPath: '',
   pwa: {
-    manifestPath: "assets/manifest.json",
+    manifestPath: 'assets/manifest.json',
```
bastienwirtz, line 12: :thinking: I think we could actually remove all the pwa configuration except from the manifestPath (used to inject the path in the header). 
The goal is to use the static manifest in the asset directory so anybody can customise the app name & other thing easily. What do you think?

bastienwirtz (changes requested):
I'm in to reduce the number of icons. I just suggested a few modifications, let me know what you think. 

Thanks again !

vue.config.js
NotWoods: [@vue/cli-plugin-pwa](https://cli.vuejs.org/core-plugins/pwa.html#configuration) looks like it generates the manifest file unless you put an absolute path in there. I'm not 100% sure, but its possible the JSON file isn't even used right now.

NotWoods:
I tweaked the Vue config so it pulls all values from the static JSON file. However, it doesn't look like you can turn off the manifest generation without using an absolute URL. I'll file a issue with Vue for that.

vue.config.js
```diff
@@ -10,50 +10,11 @@ module.exports = {
...
-      },
-      {
-        src: "./icons/safari-pinned-tab.svg",
-        sizes: "any",
-        type: "image/svg+xml",
-        purpose: "monochrome",
-      },
-    ],
+    themeColor: "#3367D6",
+    manifestOptions: require('./public/assets/manifest.json'),
```
bwirtz: Good idea!

bastienwirtz (approved):
Love idea to just require the static manifest, nicely done! Thanks!

PR #160, opened 2020-11-14
Create AdGuard Home service

simonporte:
## Description

Create an AdGuard Home service based on the one done for Pi-Hole, which displays the status of the protection.

AdGuard Home is the main alternative to Pi-Hole:
- Open Source
- Same functionalities
- But supports conditional filters (AdBlock lists), not only host lists like PiHole

## Type of change

## Checklist:

## Comments on the code

It is a copy paste of PiHole service with as few edits as possible, only 3 lines are changed.
URL is [/control/status](/control/status).
The `protection_enabled` key returns a true of false value.
I have zero knowledge of vuejs, so I had a hard time figuring why the class was not applied.
I used v-bind:class with ternary expression to be able to use a value other than the boolean returned wich was not applied.
Maybe it is not the best way to do this.

Below comparison of results showing behaviour for PiHole and AdGuard Home side by side.

### Enabled

msedge_GzDGgKbWYz

### Disabled

msedge_wj0kQd7fXZ

### Code changes

notepad++_kKWc0ccT0I

Ideas for enhancement: adding the ability to change the text from true/false to enabled/disabled.

### DIFF
--- a/src/components/services/AdGuardHome.vue
+++ b/src/components/services/AdGuardHome.vue
@@ -1,0 +1,88 @@
+<template>
+  <div>
+    <div class="card" :class="item.class">
+      <a :href="item.url" :target="item.target" rel="noreferrer">
+        <div class="card-content">
+          <div class="media">
+            <div v-if="item.logo" class="media-left">
+              <figure class="image is-48x48">
+                <img :src="item.logo" :alt="`${item.name} logo`" />
+              </figure>
+            </div>
+            <div v-if="item.icon" class="media-left">
+              <figure class="image is-48x48">
+                <i style="font-size: 35px;" :class="['fa-fw', item.icon]"></i>
+              </figure>
+            </div>
+            <div class="media-content">
+              <p class="title is-4">{{ item.name }}</p>
+              <p class="subtitle is-6">{{ item.subtitle }}</p>
+            </div>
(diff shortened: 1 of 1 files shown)


bastienwirtz (commented):
Thanks for starting to grow the custom services collection @simonporte! It looks good, I'll test it quickly asap & merge it if all good.

bastienwirtz:
@simonporte didn't know about AdGuard btw, do you use it instead of PiHole or both at the same time?

simonporte:
I'm currently testing both solutions on similar situations.
Not at the same time, because they serve the same purpose.
I was never really statisfied by pihole and ditched it long ago.
As I'm testing AdGuard Home, I give Pihole a second chance too.
My main use case is blocking ads on smartphones and TVs where adblock/ublock is not an option.

By the way, the query to AdGuard Home should only work if there is no authentication on the webui (contrary to pihole).
Any plan to support authentication at one point ?

bastienwirtz (approved):
Looks good :+1: 
I might find a way to factorize some code in the future but that fine for now. Thanks for the detailed PR description, appreciated.

bastienwirtz:
@simonporte Thanks for the feedback. Authentication on specific service would have to be implemented in each custom service card. Additional properties can be added to the service item in the `config.yaml` (like api keys) if necessary.

Issue #163, opened 2020-11-22
Search yields TypeError: Cannot read property 'key' of undefined


jo-me:
I played around with the search a bit, but it seems broken

A lot of times, it will hang after the first 2 letters and not recover from that (cancel does not work, changing search term does not do anything).
The console gives  

```
vue.runtime.esm.js:1888 TypeError: Cannot read property 'key' of undefined
    at Fi (vue.runtime.esm.js:5817)
    at k (vue.runtime.esm.js:6219)
    at _ (vue.runtime.esm.js:6319)
    at k (vue.runtime.esm.js:6193)
    at _ (vue.runtime.esm.js:6319)
    at k (vue.runtime.esm.js:6193)
    at _ (vue.runtime.esm.js:6319)
    at k (vue.runtime.esm.js:6193)
    at _ (vue.runtime.esm.js:6319)
    at k (vue.runtime.esm.js:6193)
```

As far as I can tell only the name fields are searched, but in the error case items show up that don't contain the letters.
e.g i type "po" to search for portainer, but "nextcloud" and "Home assistant" show up as well.

Any ideas?
I'm using the last release version.

jo-me:
After debugging a bit it seems to happen when several search results have the same name. 
I have two portainer instances and therefore two items here with name "portainer". I put in different subtitles for each instance.  
If I rename one to "portainer 2" the search works.

What I don't understand is why searching for "p" works in my case even though every item is found then (due to tag "app").
Only when adding the second letter it crashes for duplicate names..

PR #164, opened 2020-11-22
Fix item key for horizontal layout

jo-me:
## Description

App crashes if search results yield items with same name (in horizontal layout).
See issue for details.

Fixes  https://github.com/bastienwirtz/homer/issues/163

## Type of change

## Checklist:

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -65,8 +65,8 @@
                 {{ group.name }}
               </h2>
               <Service
-                v-for="item in group.items"
-                :key="item.url"
+                v-for="(item, index) in group.items"
+                :key="index"
                 v-bind:item="item"
                 :class="['column', `is-${12 / config.columns}`]"
               />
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -66,7 +66,7 @@
               </h2>
               <Service
                 v-for="item in group.items"
-                :key="item.name"
+                :key="item.url"
```
bastienwirtz, line 69: Just to be 100% covered, I would go for something like that (untested, but should work): 
```js
:key="index"
```
That way even duplicate urls won't be an issue

bastienwirtz (changes requested):
Good catch! Thanks for your contribution @jo-me!

src/App.vue
jo-me: You're right. That's similar to how AngularJS did it.
Fixed it & tested it.

bastienwirtz (approved):
Looks good, thanks @jo-me !

PR #166, opened 2020-11-22
Include subtitles in search

jo-me:
## Description

Currently only name and tag fields are searched.
This PR includes the subtitles as well. 
IMO, subtitles should be included because they are visible texts in the UI and it is therefore clear to the user why an item is coming up in the result list (unlike if e.g. URLs would be searchable).

## Type of change

## Checklist:


bastienwirtz (approved):
Hi @jo-me, thanks for the contribution. Look good :+1:

PR #177, opened 2021-01-07
Adds multiple pages based on different config files

### LINKED ISSUE #152: Feature Request: Enable different 'pages' via multiple configs?
Was wondering if it would be possible to augment the getConfig function to take in a parameter from the URL to possibly load a different config (and default to config.yml)? This would enable you to have navigation links within the system to other pages.

luixal:
## Description

As proposed in the related issue, this PR allows to have multiple pages based on different config files.

### How does this work?
In the links section, you can add a page with a related path, for example (look at the "Second page"):

```yml
links:
  - name: "Contribute"
    icon: "fab fa-github"
    url: "https://github.com/bastienwirtz/homer"
    target: "_blank" # optional html a tag target attribute
  - name: "Second Page"
    icon: "fas fa-file-alt"
    url: "/page2"
```

When you navigate to this link, Homer will load the `assets/config.yml` file and the `assets/page2.yml` file; then it overwrites the fields present on the second config over the first one. This way you can keep the config presets from the main config and only include the ones you want to be different in the second.

For example, if your `page2.yml` file is like this:

```yml
# this config is used by a page linked in the navbar
# this pages will use the same configuration from config.yml, but will overwrite fields present here

# this overwrites title and subtitle:
title: "Page2"
subtitle: "this is the second page"

# this overwrites message config. Setting it to empty to remove message from this page and keep it only in the main one:
message:

# as we want to include a differente link here (so we can get back to home page), we need to replicate all links or they will be revome when overwriting the links field:
links:
  - name: "Home"
    icon: "fas fa-home"
    url: "/"
  - name: "Contribute"
    icon: "fab fa-github"
    url: "https://github.com/bastienwirtz/homer"
    target: "_blank" # optional html a tag target attribute
  - name: "Wiki"
    icon: "fas fa-book"
    url: "https://www.wikipedia.org/"

# we keep the first group from the main page, but remove the second group. We need to replicate that first group or it will be removed:
services:
  - name: "NEW"
    icon: "fas fa-cloud"
    items:
      - name: "Awesome app"
        logo: "assets/tools/sample.png"
        subtitle: "Bookmark example"
        tag: "app"
        url: "https://www.reddit.com/r/selfhosted/"
        target: "_blank"
      - name: "Another one"
        logo: "assets/tools/sample2.png"
        subtitle: "Another application"
        tag: "app"
        url: "#"
```

It will overwrite fields `title`, `subtitle`, `message`(gets removed as it's empty), `links`and `services` with config from this file, but kill keep other fields from the default `config.yml` file.

### Considerations

For this to work without errors, I had to change line (36@ConnectivityChecker.vue)[https://github.com/bastienwirtz/homer/blob/3786f80dae2df7780d19ba8ffd9374ef3c2fc30f/src/components/ConnectivityChecker.vue#L36] so it refers to `window.location.origin` instead of `window.location.href'  or it would try to use `/page2?alive` for checking connectivity. Seems to work fine this way.

Fixes #152 

## Type of change

## Checklist:


bastienwirtz:
Hi @luixal!

Thanks for your contribution! Same here, that sounds great, I will look more closely into it soon!

src/App.vue
```diff
@@ -153,6 +153,13 @@ export default {
     let config;
     try {
       config = await this.getConfig();
+      const path = (window.location.pathname != '/') ? window.location.pathname : null;
```
pdevq, line 156: Instead of using a path, maybe use a query string instead, which allows for subdirectory hosting.
```suggestion
      const path = (window.location.search.substring(1) != '') ? window.location.search.substring(1) : null;
```

src/App.vue
```diff
@@ -153,6 +153,13 @@ export default {
...
+      if (path) {
+        let pathConfig = await this.getConfig(`assets${path}.yml`); // the slash (/) is included in the pathname
```
pdevq, line 158: When using a query string the slash is no longer part of the pathname and needs to be added.
```suggestion
        let pathConfig = await this.getConfig(`assets/${path}.yml`);
```

src/App.vue
```diff
@@ -13,7 +13,7 @@
       <section v-if="config.header" class="first-line">
         <div v-cloak class="container">
           <div class="logo">
-            <img v-if="config.logo" :src="config.logo" alt="dashboard logo" />
+            <a href="/"><img v-if="config.logo" :src="config.logo" alt="dashboard logo" /></a>
```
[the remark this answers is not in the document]
pdevq: If using a query string, just clear it.
```suggestion
            <a href="?"><img v-if="config.logo" :src="config.logo" alt="dashboard logo" /></a>
```

src/components/ConnectivityChecker.vue
```diff
@@ -33,7 +33,7 @@ export default {
   methods: {
     checkOffline: function () {
       let that = this;
-      return fetch(window.location.href + "?alive", {
+      return fetch(window.location.origin + "?alive", {
```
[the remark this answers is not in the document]
pdevq: With a query string this can reverted back to `href`.
```suggestion
      return fetch(window.location.href + "?alive", {
```

docs/configuration.md
```diff
@@ -81,6 +81,11 @@ links:
   - name: "link 2"
     icon: "fas fa-book"
     url: "https://github.com/bastienwirtz/homer"
+  # this will link to a second homer page that will load config from page2.yml and keep default config values as in config.yml file
+  # see url field and assets/page.yml used in this example:
+  - name: "Second Page"
+    icon: "fas fa-file-alt"
+    url: "/page2"
```
pdevq, line 88:
```suggestion
    url: "?page2"
```

pdevq (commented):
This introduces a breaking change for #8 (subdir hosting). The quickest fix I could think of is using a query string instead of a path for additional pages. I've added the corresponding suggestions in the review.

Personally using `#page2` instead of `?page2` in the config seems more visually pleasing (and intuitive) to me, but as anchor links don't reload the page an additional mechanism would be needed.

On a side note and please correct me if I'm wrong, even ignoring the breaking change, doesn't a path like `/page2` require additional configuration on the server? This would than also not work on the current docker image. For this to work on `nginx` I had to configure:
```nginx
location /page2 {
    alias /usr/share/nginx/html;
    try_files $uri $uri/index.html $uri/ =404;
}
```
More importantly one has to be careful that the server does not add a trailing slash, which breaks the config file retrieval.

luixal:
Changes sound good to me. Didn't notice the #8 feature. Maybe a doc with all features will be nice to have :)

src/App.vue
pdevq, line 16:
```suggestion
            <a href="#"><img v-if="config.logo" :src="config.logo" alt="dashboard logo" /></a>
```

src/App.vue
pdevq, line 156:
```suggestion
      const path = (window.location.hash.substring(1) != '') ? window.location.hash.substring(1) : null;
```

src/App.vue
pdevq, line 158:
```suggestion
        let pathConfig = await this.getConfig(`assets/${path}.yml`);
```

src/components/ConnectivityChecker.vue
pdevq, line 36:
```suggestion
      return fetch(window.location.href + "?alive", {
```

src/App.vue
```diff
@@ -153,6 +153,13 @@ export default {
     let config;
```
pdevq, line 153: Just reload the page on an hash change event.
```suggestion
    let config;
    window.onhashchange = function() { location.reload(); };
```

pdevq (commented):
To be fair, I only checked for the feature after the fact. Still running an old fork locally that does not support it. I've just had my fair share of fights with applications that don't play well with relative paths and noticed that this change may be problematic in a subdir deployment. Some features are listed under `Features` in the `README.md`, maybe hosting in a subdir could be added?

I'm still more fond of using `#page2` rather than `?page2`. Not just visually, but also because using the query string could block or make it harder for future additions to make use of it. I've added another review with a quick implementation for `#page2`.

luixal:
Just tested your suggested changes and it works great. Just cleanning up and will push for merge. Thanks! :)

bastienwirtz:
Thanks for the great work @luixal & @pdevq !

I merged that as is to just [modified a few thing](https://github.com/bastienwirtz/homer/commit/ba07da6b1011e77c9ed42e8643e62b903c6c6d7b): 
- Little clean (commented code, lint, ...)
- Renamed `page2.yml` to `another-page.yml.dist` (important thing is the `.dist`, all "templates" config in this repo follows this convention) & added a bit more description. 
- Made possible to switch page without a full reload!

Hope that sounds good to you!

btw, I updated the [demo](https://homer-demo.netlify.app/) to show the second page feature.
Thanks again!

pdevq:
Awesome work! The 'no full reload' is a lot better. Guess I'll have to finally update my local copy, don't know yet what I need the multipages for, but I really like them.

luixal:
Great! Avoiding full reload was a next stage I had on mind but I didn't found the moment to do it, awesome you could!

PR #190, opened 2021-02-13
README: Generate static files

### LINKED ISSUE #189: How to build HTML files?
Hello, I just discovered this project and I'm wondering if it's possible to use Homer to build HTML files?

Something like this:

```
docker run -v $PWD/assets/:/www/assets -v $PWD/output/:/www/output b4bz/homer:latest build
```

It would compile the files then stop, and assets would be available in the `output/` directory, then my web server could serve them, without having to run the Homer server.

alexislefebvre:
## Description

Explain how to build static files without using the built-in server.

Fixes #189

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ See [documentation](docs/configuration.md) for information about the configurati
 
 ### Using docker
 
+#### Using built-in server
+
 To launch container:
 
 ```sh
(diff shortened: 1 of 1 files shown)


bastienwirtz:
Thanks for the contribution @alexislefebvre!
Even if it might be more convenient, I wouldn't recommend building the app that way, as I mentioned in the related issue: https://github.com/bastienwirtz/homer/issues/189, it's totally fine to do so, but I wouldn't add it to the documentation (Homer could be installed & used in a dozen ways, but I just want to document the common way to keep thing simple).

alexislefebvre:
Right, no problem.

PR #214, opened 2021-04-25
Add crossorigin use-credentials attribute to manifest tag

pkoenig10:
## Description

When serving Homer behind an authentication proxy (ex. https://github.com/pusher/oauth2_proxy) the manifest request will fail because cookies are not sent with the manifest request.

See https://developer.mozilla.org/en-US/docs/Web/Manifest:
> Note: If the manifest requires credentials to fetch - the `crossorigin` attribute must be set to `"use-credentials"`, even if the manifest file is in the same origin as the current page.

## Type of change

## Checklist:

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -12,6 +12,7 @@ module.exports = {
   publicPath: "",
   pwa: {
     manifestPath: "assets/manifest.json",
+    manifestCrossorigin: "use-credentials",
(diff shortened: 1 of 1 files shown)


bastienwirtz (approved):
Thanks for your contribution @pkoenig10! Looks good :+1:

PR #224, opened 2021-05-27
Add gitpod support

BCNelson:
Gitpod is a browser based development environment  that I have found super useful in reviewing pr's and lowering the barrier to entry for new contributors.


bastienwirtz:
Hi there, sorry for not looking into it sooner. Thanks for your contribution. I think it's a bit specific and I try to keep Homer as light & simple as possible, so I'm not super interested to integrate it to the code base. 
Thanks anyway :)

Issue #208, opened 2021-04-13
include in yaml


ostgardh:
Is it possible to have include in yaml?
Eg.

message: !include message.yaml

Then is Nice to have same message on all pages..

bastienwirtz:
Hello, 

Unfortunately it's not possible, it's not in the yaml spec, and it's not a custom feature from the lib either.

PR #237, opened 2021-06-20
Overrides/hides services based on detected domain

acmelaw:
## Description

This PR introduces an optional feature to detect what "domain" the user is on (e.g. local, remote, vps, etc.) based on user-configured rules, and allows overriding properties & hiding services based on this.

For example, the domain can be detected as "local" if browser address bar hostname matches either of "127.0.0.1", "localhost", or if "http://127.0.0.1:8080" pings successfully.

Screenshot-20210620224930-479x185

The following are implemented:
- Rules to detect domain based on:
  - Browser address bar URL (detects domain from user-defined hostname list, or full URL via regex)
  - URL pinging (matches domain group based on URL list ping success)
- Show the detected domain on top-right corner
- Flag to show/hide services
- Override service properties if domain matched

Configuration additions are described in [docs/configuration.md](docs/configuration.md).

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -24,6 +24,26 @@ footer: '<p>Created with <span class="has-text-danger">❤️</span> with <a hre
 columns: "3" # "auto" or number (must be a factor of 12: 1, 2, 3, 4, 6, 12)
 connectivityCheck: true # whether you want to display a message when the apps are not accessible anymore (VPN disconnected for example)
 
+domainChecker: # optional domain checker
+  enabled: true
+  method: "url" # 'url' (matches browser URL) or 'ping' (pings specified URL)
+  url:
+    method: "hostname" # 'hostname' (matches hostname) or 'regex' (matches href via regex)
+    hostnames:
+      local:
+        - "localhost"
+        - "127.0.0.1"
+      remote:
+        - "github.io"
+    regex:
+      http: "^http"
+      https: "^https"
+  ping: # pings url(s) and matches domain
+    local:
+      - "http://localhost:8080"
+    remote:
+      - "github.io"
+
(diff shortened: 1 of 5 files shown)


khaledhassan:
+1 for a feature like this!

My use-case would be to change the URLs of services that are available over a local network and a VPN like Tailscale/Nebula/etc., given that not all services might not be accessible over the VPN/it might be faster to have a direct connection if on the same network/DNS names may not be set up or accessible/etc. Being able to hide the services that are unavailable based on connection method is a cherry on top!

My original plan for handling such a scenario would be to have an `additionnal-page.yml.dist`-like second tab where I duplicate the services in the config file (changing the URLs as necessary), but maintaining two copies of the config didn't feel that clean. I then tried the YAML anchors described in the `docs/tips-and-tricks.md` but the YAML parser doesn't (seem to (I haven't really dug into the code to know if this is the case)) remember the fields defined in the `config.yml` file as it parses the other page's config. Lastly, I would have hoped that a `!include`-like feature was available (like in #208), but alas.

Maybe I'll try building from source with your changes included. Hope it gets merged -- certainly better than the path I was headed down!

PedroC1999:
What additional work would be required for this to pass the PR?

hcoohb:
This would be a really great addition!

bastienwirtz:
Hi there ! Sorry for not looking into it sooner. 
Thanks for your contribution! That's an interesting use case, but I try to keep homer as simple and dry as possible, and I can't integrate all possible use case. I would recommend to use the multi page feature (different configs can easily be maintained with variables and templating, with Ansible for example), or even use multiple Homer instance. 
Thanks you!

PR #239, opened 2021-06-24
fix card border radius when it's the only child

azrikahar:
### Context
When there's only one card, the css rule `body .layout-vertical .column div:last-child .card` (line 266 of `src/assets/app.scss`) kicks in as it is indeed the "last-child" in this scenario. However, the css property `border-radius: 0 0 5px 5px;` caused the `top-left` and `top-right` border radius to reset to 0, which shouldn't be the case for a single child card should have all rounded corners.

Screenshot of issue & devtools to show the css selector involved:
chrome_Q6OnL6lO7T

Just to ensure this is not caused by my custom css in the screenshot (which just changes the color variables provided & font), we can also see the same behavior over at the demo page here: https://homer-demo.netlify.app/#additionnal-page

### Summary of change
When I was checking the stylesheet, I noticed there's an approach using `border-top-left-radius` etc to make sure we don't reset other corners to 0, and apparently it came from the bulma css as seen here: 
chrome_oV1wHgyW9N

By mimicking the same approach, I then changed `body .layout-vertical .column div:last-child .card` and `body .layout-vertical .column div:first-of-type .card` to fix the issue, and here's an after screenshot (with list 2 being unaffected at all):
chrome_MZzl1bkS9G

Lastly since Bulma also has `border-radius: 0.25rem` by default for the `.card` class, I removed it at line 227 so that it will be uniformly 0.25rem.

## Type of change

## Checklist:

### DIFF
--- a/src/assets/app.scss
+++ b/src/assets/app.scss
@@ -224,7 +224,6 @@ body {
   }
 
   .card {
-    border-radius: 5px;
     border: none;
     box-shadow: 0 2px 15px 0 rgba(0, 0, 0, 0.1);
     transition: cubic-bezier(0.165, 0.84, 0.44, 1) 300ms;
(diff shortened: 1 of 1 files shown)


bastienwirtz:
Good catch @azrikahar !! 
Thanks for the great PR description and the screenshot. This is very much appreciated.

PR #255, opened 2021-08-05
add cors ping

stubbfel:
## Description

add a new service : `CorsPing`. This new service allow the user to use a cors proxy server.

Fixes # (issue)

## Type of change

## Checklist:

- [x] I have made corresponding changes the documentation (docs/customservices.md).

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -51,3 +51,70 @@ For Paperless you need an API-Key which you have to store at the item in the fie
 ## Ping
 
 For Paperless you need an API-Key which you have to store at the item in the field `apikey`.
+
+### CORS Issues
+
+If try you to "ping" services, which have a differ domain as the homer service then may you get a `Cross-Origin Request Blocked` messages (https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors).
+
+Possible solutions:
+
+* Modify the target sever configuration so that the response of the server included following header- `Access-Control-Allow-Origin: *` (https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests)
+
+* Use a cors proxy sever  like `cors-container` (https://github.com/imjacobclark/cors-container)
+
+#### Example docker compose file
+
+```yml
+---
+version: "2"
+services:
+  homer:
+    image: b4bz/homer
+    #To build from source, comment previous line and uncomment below
+    #build: .
+    container_name: homer
+    volumes:
+      - /your/local/assets/:/www/assets
+    ports:
+      - 8080:8080
+    #environment:
+    #  - UID=1000
+    #  - GID=1000
+    restart: unless-stopped
+  #The cors-container service which avoid "allow" cross-domain requests, can use with the CorsPing type
+  cors-container:
+    image: imjacobclark/cors-container:latest
+    container_name: cors-container
+    ports:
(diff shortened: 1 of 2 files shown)


zanna-37:
Why not just adding `mode: "no-cors"`? Am I missing something?
`fetch(url, { method: "HEAD", cache: "no-cache", mode: "no-cors" })`

stubbfel:
hi @zanna-37 , 
i tried your approach.
i tested with firefox and chromium, but it seems not to work :(

but if you would find a way to get it work, would be great.

FYI: some more information about "no-cores" https://stackoverflow.com/questions/43262121/trying-to-use-fetch-and-pass-in-mode-no-cors/43268098#43268098

zanna-37:
Thanks a lot for you patience and also for the good resource! I see your point, with `mode: "no-cors"`. You can distinguish a server that responds from one that doesn't (connection reset etc), but if the server responds with an error, it is impossible to distinguish that from a `200 OK`.

Anyway, I like you approach and I hope it can be merged soon.

One small proposal, is it possible to modify the current Ping to add your changes, instead of creating a new one? I think it would ease the merging process.
In particular I'm thinking about adding a `health_url` property that js uses to check the server. If the property is not set it defaults to the server url.
What do you think?

stubbfel:
yes it would be another way to solve this issue, but i would prefer `proxy_url` instead of `health_url`.

may @bastienwirtz  should decide which way should be taken:

1. add new service type (CorsPing)
2. enhance ping service type by a `proxy_url` field

zanna-37:
Yeah, I'm also interested in @bastienwirtz opinion.
In the mean time, why do you prefer the `proxy_url` version?
It fits very well with you scenario (with the cors proxy), but a was imagining a more general solution. For example, a server may expose a path to check its health (obviously setting the right headers to allow cors). Therefore the `health_url`.
What do you think?

stubbfel:
Ok, i see . Yes then `health_url` could be a better name.

stubbfel:
@zanna-37  @bastienwirtz  see pr #267  for the `health_url` solution

bastienwirtz (commented):
Hi @stubbfel, 
Thanks for helping with the CORS issue ! definitely impact a lot of users. 
It's more of an hosting issue than really related tohomer itself, so I created a [troubleshooting](https://github.com/bastienwirtz/homer/blob/main/docs/troubleshooting.md) section. 

Regarding  the cors ping component, I think it's not really a solution because this impact all of the custom services, not just the ping one. And the `url` config in the base service can be use to point on a cors proxy. So I prefere to not add it. 

Feel free to improve the troubleshooting section if you feel like it. 
Thanks a lot!

bastienwirtz:
Hey @stubbfel ! 
I added a property `endpoint` in addition to `url` to make possible to have an alternative, and made it available for all custom service (not only ping) by  creating a [`mixin`](https://github.com/bastienwirtz/homer/blob/main/src/mixins/service.js). Checkout the [updated documentation](https://github.com/bastienwirtz/homer/blob/main/docs/customservices.md#common-options) and the [reworked ping service](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Ping.vue).

I hope this works for you, let me know what you think!

stubbfel:
@bastienwirtz I'm fine with this.I think that should work, for the most usecases.

PR #267, opened 2021-08-31
add health_url

stubbfel:
## Description

@zanna-37 mention in pr  #255 to extend the ping service by a `health_url` instead of add a new  service type `CorsPing`.

Now the ping will check if the `health_url` is set. If it is, then it use  the value  of field `health_url` for the fetch call. Otherwise  the `url` field will still be used.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -47,7 +47,80 @@ The Radarr/Sonarr API key can be found in Settings > General. It is needed to ac
 
 For Paperless you need an API-Key which you have to store at the item in the field `apikey`.
 
-
 ## Ping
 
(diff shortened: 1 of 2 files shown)


bastienwirtz (commented):
Following on your 1st PR, I don't think this is necessary as the `url` config can be replaced by the proxied URL.

zanna-37:
I strongly disagree. The `url` is the url of the service, and it does not need anything special. Adding a `health_url` allows to decouple the service from with the hearth-beat function.
One can have an API to signal the health of the service (with all the headers set) while leaving the service url unmodified.

This approach is an optional addition, and nothing changes for existing users.

PR #274, opened 2021-09-11
Added custom hotkey support

### LINKED ISSUE #110: [Feature Request] change keyboard hotkeys
Hello!

Because of / is hard to reach on a german keyboard (its shift + 7) it would be nice to see the feature to custom assign the seach hotkeys

BR

robinschneider:
## Description

Fixes #110 

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -19,6 +19,8 @@ logo: "assets/logo.png"
 # icon: "fas fa-skull-crossbones"
 
 header: true # Set to false to hide the header
+# Optional: Different hotkey for search, defaults to "/"
+# hotkey: "Shift"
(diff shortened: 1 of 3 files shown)


robinschneider:
This PR allows to change the hotkey for search. 
The default hotkey is `/` which is not usefull on german keyboard layout.
The change allows to override the hotkey in the configuration.
Setting a hotkey is optional and not required.

robinschneider:
The `/` is the default value in `SearchInput.vue`, so I thing it should be fine with `config.hot key?.search?`. That way there is no need to set a hotkey and the issue should be fixed as well.

robinschneider:
So I fixed the issue, but both `:hotkey="config.hotkey?.search"` as well as `:hotkey="config.hotkey?.search?"` did not work, so I used a function to check it.
Now there is no need to set a default value in the `config.yml` as `/` is always the default.
If set, the default will be ignored and the custom hotkey used.

PR #238, opened 2021-06-21
Added feature for subfolder hosting

boerniee:
By setting the SUBFOLDER env var, you can host Homer in a subfolder behind a reverse proxy.

## Description

I would like to host Homer behind a reverse proxy (traefik) in a subfolder. With using the StripPrefix middleware, the index page is loading but couldnt load any assets, because they had still absolute urls. To get this working, I have added a little modification to the entrypoint.sh script, which adds a symbolic link to the /www directory. This symbolic link has the name of the subfolder and points to the /www directory itself. So the application is available under yourdomain.com/ and yourdomain.com/<nameofyoursubfolder>/.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I've read & comply with the [contributing guidelines](https://github.com/bastienwirtz/homer/blob/main/CONTRIBUTING.md)
- [x] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. 
- [x] I have made corresponding changes the documentation (README.md).
- [x] I've checked my modifications for any breaking changes, especially in the `config.yml` file

PR #292, opened 2021-09-23
add url prefix with doc

### LINKED ISSUE #290: [FR] Add URL prefix
Hey,
is there any possibility to add a URL prefix for running homer behind a reverse-proxy with a given URL eg.

example.com/dashboard

I tried it with traefik but had no luck. Got a 404 or a blank page (without and with prefix-stripping).
I looked up the source and found some occurences of BASE_URL environment variable.
Could this be a solution?

Thanks

mad73923:
## Description

Add URL prefix capabilities for reverse proxy applications

Fixes #290 

## Type of change

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,15 @@ environment:
   - GID=1000
 ```
 
+In case homer runs behind a reverse proxy (eg Traefik) and a URL prefix is used, this prefix can be set via the environment variables: (default: /)
+
+```yaml
+environment:
+  - HOMER_URL_PREFIX=/homer
+```
+
+By setting the prefix, homer can be accessed via eg www.example.com/homer
+
 ### Using the release tarball (prebuilt, ready to use)
 
 Download and extract the latest release (`homer.zip`) from the [release page](https://github.com/bastienwirtz/homer/releases), rename the `assets/config.yml.dist` file to `assets/config.yml`, and put it behind a webserver.
--- a/vue.config.js
+++ b/vue.config.js
@@ -10,6 +10,7 @@ module.exports = {
       .end();
   },
   publicPath: "",
+  outputDir: "dist/" + (process.env.HOMER_URL_PREFIX || ""),
   pwa: {
     manifestPath: "assets/manifest.json",
     manifestCrossorigin: "use-credentials",


mad73923:
similar to #238

fbartels:
Homer works just fine from a subfolder even without your change. The linked pr also only changes the subfolder for the container.

mad73923:
sure it does. but I couldn't find a other way to achieve this inside a container. The linked PR is working for docker, but only for that.
There might be folks out there running homer locally behind a reverse proxy.

This solution covers both, containerized AND local deployment.

fbartels:
> This solution covers both, containerized AND local deployment.

That may be true, but you change is only really relevant when compiling Homer yourself and even if you don't, you could still move the build result in any directory that you want afterwards.

mad73923:
Firstly, homer gets compiled every time it is deployed in a container. Why not using the move functionality in this step?
Secondly, moving things inside a container manually is a bad way to deploy! You'd have to do this with every update / deployment again and again.
Additionally, this change doesn't change anything to the legacy users. Only if the environment variable is set, things will change.

bastienwirtz:
@fbartels In docker, the subdirectory will not work because If you set a subpath on your reverse proxy, Homer will not be aware of it and try to requests the the resources on `/`.

@mad73923 I like the idea, and this is more elegant than the #238 solution, but it works only if you build the docker image yourself right? If you pull it from dockerhub (like most peolple do), the subfolder env var will not be taken into account. In #238, it works without image rebuild. 

That being said, maybe the best solution would be to add a `proxy_path` or `homer_path` or something like that in the `config.yml` configuration and prefix resources requests with it. And for the hosting, leave it to the vhost or proxy configuration (so no modification necessary on the build path).

What do you think @mad73923 @boerniee @fbartels ?

boerniee:
Yes, this one is more elegant, but it doesnt work for pulled images from Dockerhub. As you @bastienwirtz mentioned, I also think the most users will do this.
For people which host homer locally, the effect of building the project and being placed in the right dir automatically, or downloading the prebuild project and placing it manually in the right location, is the same. Here I also think the most people will download the prebuild project. In my opinion there is no need to change anything for people who locally host homer, because instead of placing it in the www-root they can move it into another folder.

Another thing I would like to mention is, that I like the idea of maintaining the subfolder config in the docker labels side by side with the reverse-proxy configuration for that container.
Maybe this thing could be solved by doing some darkhttp config?

mad73923:
Hey @bastienwirtz and @boerniee ,
you're right. The container will already be compiled when pulling. I forgot about that.
Unfortunately, after a quick glimpse at darkhttp, ther's no prefix parameter available.
Do you have any better ideas?

mad73923:
since this feature was implemented in #238 i close this PR

PR #302, opened 2021-10-02
- added wikijs service to display a new update is available if the wi…

tobidemski:
## Description

Adds the wikijs service to display "A new version is available!" as a subtitle and a orange yellow dot with the new version if a new update is available otherwise the dot is green and the current version is displayed.

** **INFO** **
Maybe we should wait with this pull request until pr #301 is merged to inlude his new method "fetchOptions" in my fetch call?

## Type of change

## Checklist:


D4rkiii:
Hey @bastienwirtz,

i updated my pull request so far but im struggle a bit with the "subtitle" which i set if none is provided by the user.
I set the value after the service call with :

this.item.subtitle = "update available"

Im not sure how to resolve this but you have a check in the generic service which sets the "no-subtitle" css class on the container and center the content. But if i set the subtitle by my self, the indicator isn't anymore on top right instead it is right and centered.
Im not familiar with vue and it would be nice if you could take a look at my current pr'd state.
Otherwise I would be ready with my pr and i tried to fix all your mentioned parts.

Looking for your respond!

PR #311, opened 2021-10-12
Fix #121 - Change default theme and layout from config

Aryess:
## Description

This PR adds default values in the config for the theme (dark/light/auto), and the layout (column/list). Respects the current default, if no config is changed, default will still be auto/column.

Fixes #121

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] I've read & comply with the [contributing guidelines](https://github.com/bastienwirtz/homer/blob/main/CONTRIBUTING.md)
- [X] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. 
- [X] I have made corresponding changes to the documentation (README.md).
- [X] I've checked my modifications for any breaking changes, especially in the `config.yml` file

PR #313, opened 2021-10-12
Fix #312 Vue warning in dev mode when having multiple service groups in list mode

### LINKED ISSUE #312: VueJS warning when rendering multiple service groups in list mode
Because services uses their sub-group index as a key, and are all displayed in the same div, if there are multiple groups, some services end up having the same key.

Screenshot from 2021-10-12 11-49-08

This is only a warning, and doesn't currently impact rendering, but still worth fixing.

Aryess:
## Description

Something I've noticed while working on #311, if multiple service groups are declared, Vue will raise warnings when rendering them in list mode, as they are using they service group index as a key, while all being rendered in the same div (which is why it's only an issue in list mode, in column they are in separate divs). This fix changes the key to use the service group name, as the alternatives would be to isolate each group which would require way more changes to keep things displaying the same.

Before:
Screenshot from 2021-10-12 11-59-50

After:
Screenshot from 2021-10-12 12-00-15

Fixes #312

## Type of change

## Checklist:

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -73,7 +73,7 @@
               </h2>
               <Service
                 v-for="(item, index) in group.items"
-                :key="index"
+                :key="group.name + '-' + index"
                 :item="item"
                 :proxy="config.proxy"
                 :class="['column', `is-${12 / config.columns}`]"


bastienwirtz:
Hey @Aryess, 
sorry for not looking into that earlier :see_no_evil:, this has been fixed since then. 
Thanks anyway for contributing.

Aryess:
No worries :)

PR #344, opened 2021-12-20
Warn about exposed apikeys

michaelkrieger:
## Description

The config.yml is loaded by the client at /assets/config.yml and a request made to the "serverurl?apikey=..." for custom services.

All of this is fine, but is should be clearly pointed out that the homer dashboard when including your apikey should not be exposed to untrusted users.  This seems like a documentation issue which could easily be added.

Fixes # (issue)

## Type of change

## Checklist:


bastienwirtz (approved):
Hey @michaelkrieger, Very good point. Thanks for your contribution !

Issue #261, opened 2021-08-20
Update docker image to be able to run as non-root user


justsomebody42:
Hi,

I'm running 21.07.1 in a Kubernetes cluster.
It works super smooth and I'm very happy with the application itself.

I would like to run the image as non-root user.
Are there plans to improve the security of the docker image?

My current Deployment looks like this and I'd be happy, if I could turn on runAsNonRoot and enable runAsUser instead of relying on the endpoint script to finally run the server with the provided userid :)

While I can understand that it is more end user friendly to repair file permissions and ownership during the startup, I think focussing on security and reminding users to prepare their environment accordingly would be beneficial during runtime.

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: homer
  namespace: homer
spec:
  replicas: 1
  selector:
    matchLabels:
      app: homer
  template:
    metadata:
      labels:
        app: homer
    spec:
      #      securityContext:
      #        runAsNonRoot: true
      volumes:
        - name: homer
          persistentVolumeClaim:
            claimName: homer
      containers:
        - name: homer
          image: b4bz/homer:21.07.1
          securityContext:
            privileged: false
            allowPrivilegeEscalation: false
            readOnlyRootFilesystem: true
          #            runAsUser: 1032
          #            runAsGroup: 65537
          env:
            - name: UID
              value: "1032"
            - name: GID
              value: "65537"
          volumeMounts:
            - name: homer
              mountPath: /www/assets
          ports:
            - containerPort: 8080
          resources:
            requests:
              memory: "64Mi"
              cpu: "250m"
            limits:
              memory: "128Mi"
              cpu: "500m"

```

grantland:
Setting `UID` and `GID` environment variables seems to work for me:

https://github.com/bastienwirtz/homer/blob/b6c667a129867db8c0f8aa29d12013b47e0c6516/docker-compose.yml#L13-L15

justsomebody42:
Hi,

The entrypoint.sh file uses/requires root permissions to do the setup actions.
`chown -R $UID:$GID /www/assets` fails, when the container itself is started without root permissions.
Currently the script relies on having root permissions and afterwards starts the process with the GID/UID retrieved from the environment.

Using Kubernetes, you can go a step further and start the whole container without root permissions, which makes it a lot safer.
This would require the image to be prepared in a way that root permissions are not required anymore.

Maybe it's possible to move the setup commands into the Dockerfile?
[...]

PR #375, opened 2022-02-14
Fix(entrypoint.sh): Skip chown & su-exec when running as non-root | Fixes: #325 & #261

### LINKED ISSUE #325: Run entire container as non-root user not possible because of chown in entrypoint.sh script
I have issues with line 14 and 15 in entrypoint.sh because I try to change the user of the docker container entirely and that is currently not possible without some strange hackarounds. See the following commands and results with a default setup:

`docker exec home whoami`
`root`

`docker exec home top`
````
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
   35     0 root     R     1580   0%   1   0% top
    1     0 homer    S      784   0%   0   0% darkhttpd /www/ --no-listing --por
````

So the darkhttpd process is running as homer user but the the container isn't. I tried simply this to build this Dockefile to make the whole container run as another user.

````
FROM b4bz/homer:latest

ENV USER homer
ENV UID 8000
ENV GROUP homer
ENV GID 8000

RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP}
USER homer
````

But the last line will fail because ./entrypoint.sh contains chmod entries that will be executed by the homer user instead of the root user. I think this should be done earlier in the Dockerfile of this project rather than in an sh script to prevent this.

### PR DESCRIPTION
**Pull request by Nold360**
## Description

This is a simple workaround in the `entrypoint.sh` which makes homer fully working when running as non-root e.g. on kubernetes.

It makes homer compatible with the full set of [pod]securityContexts:

``` yaml
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 911
  runAsGroup: 911
  fsGroup: 911

securityContext:
  privileged: false
  allowPrivilegeEscalation: false
  readOnlyRootFilesystem: true
  runAsUser: 911
  runAsGroup: 911
```

Fixes #325 & #261 

## Type of change

## Checklist:

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -11,5 +11,10 @@ fi
 # Install default config if no one is available.
 yes n | cp -i /www/default-assets/config.yml.dist /www/assets/config.yml &> /dev/null
 
-chown -R $UID:$GID /www/assets
-exec su-exec $UID:$GID darkhttpd /www/ --no-listing --port "$PORT"
+# Skip chown & su-exec if we are already non-root
+if [ "$(whoami)" == "root" ] ; then
+    chown -R $UID:$GID /www/assets
+    exec su-exec $UID:$GID darkhttpd /www/ --no-listing --port "$PORT"
+else
+    exec darkhttpd /www/ --no-listing --port "$PORT"
+fi


lavinj87:
@Nold360 have you published an image with this fix? Instead of waiting

Nold360:
> @Nold360 have you published an image with this fix? Instead of waiting

@lavinj87 here you go: `nold360/homer:noroot`

werwolfby:
Thank you very much for the fix! 

Unfortunately the image it doesn't support ARM. Will do my own soon.

werwolfby:
I've build my own image with all platforms: `amd64`, `arm32/v7` and `arm64`: https://hub.docker.com/repository/docker/werwolfby/homer/tags?page=1&ordering=last_updated 

```bash
docker pull werwolfby/homer:22.02.1-noroot
```

bastienwirtz:
Hi there !
Thanks for contributing. The docker file has been entirely reworked to run as a non-root user, it should solve this. Feel free to give your feedback!

PR #393, opened 2022-03-05
Check if path has data before adding a trailing slash (/)

### LINKED ISSUE #374: Ping service adds trailing slash to endpoint URL
Using this example config:
```
- name: "Foo"
  type: "Ping"
  url: "https://foo.tld"
  endpoint: "https://foo.tld/health?token=XXXXX"
```

I can see homer sends a request to `https://foo.tld/health?token=XXXXX/`, causing the service to parse the token as `XXXXX/` and return 403.

Tested on version 22.02.1.

jamesmacwhite:
## Description

This modifies the fetch method to check if the path variable has data before adding a trailing slash. Currently by not doing this services using the Ping type always have a trailing slash added to the end of the specified `endpoint` or `url` value. Depending on if the endpoint expects this, this can break the ping check.

The current return value is assuming that `${path}` is always present, this is not always the case or at least isn't with the Ping type, as demonstrated with reported issue/example from the linked GitHub issue.

This change will ensure the contents of `${path}` is added when it's present, otherwise it won't leave a trailing slash in the return value when it is not needed.

Fixes #374

## Type of change

## Checklist:

- ~~I have made corresponding changes to the documentation (README.md).~~

### DIFF
--- a/src/mixins/service.js
+++ b/src/mixins/service.js
@@ -31,7 +31,13 @@ export default {
         path = path.slice(1);
       }
 
-      return fetch(`${this.endpoint}/${path}`, options).then((response) => {
+      let url = this.endpoint;
+
+      if (path) {
+        url = `${this.endpoint}/${path}`;
+      }
+
+      return fetch(url, options).then((response) => {
         if (!response.ok) {
           throw new Error("Not 2xx response");
         }


bastienwirtz (approved):
Hi @jamesmacwhite, nice catch, thanks for your contribution!

PR #400, opened 2022-03-17
Add note for using fs version 5.15.4

Little-Karl:
## Description

Not much just one line of note. Remember to use version 5 on the font awesome website for icons to work. Maybe I'm just stupid but took me quite a while to figure it out. It's my first time using homer

### DIFF
--- a/public/assets/config.yml.dist
+++ b/public/assets/config.yml.dist
@@ -1,6 +1,7 @@
 ---
 # Homepage configuration
 # See https://fontawesome.com/icons for icons options
+# Use version 5.15.4
 
 title: "Demo dashboard"
 subtitle: "Homer"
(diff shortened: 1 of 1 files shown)


public/assets/config.yml.dist
```diff
@@ -1,6 +1,7 @@
 ---
 # Homepage configuration
 # See https://fontawesome.com/icons for icons options
+# Use version 5.15.4
```
Roundaround, line 4: Perhaps instead of adding an extra line, could you change the url to point to the v5 list?

https://fontawesome.com/v5/search

Roundaround (changes requested):

public/assets/config.yml.dist
Little-Karl: I accidentally deleted the 2 old patches without know it will also remove the pull request. At lao to last one didn't change the documentations and docker rework. So I made 2 pulls #404 #405. Change on main branch and docker-rework branch

PR #401, opened 2022-03-19
Portainer service - Fix for down endpoints

Zareix:
## Description

When Portainer has multiples endpoints, if one was down then the service would crash (the variable endpointContainers being 'undefined'). This fix just add a condition to add the endpoint containers if they exist/are up.

## Type of change

## Checklist:


Roundaround (approved):
Good catch, thanks!

PR #402, opened 2022-03-19
Portainer service - Select environments

Zareix:
## Description

Added an entry to the .yml config file for portainer called "environments", in order to select specific connected portainer environments.

## Type of change

## Checklist:


Zareix:
It might be useful only on some really specific use cases, but since it's only a few lines of code, I suggest it anyway.

For example in my situation : I have only one instance of portainer and homer, althought I have 2 portainer environments (2 devices). With this feature, I can have a portainer link for each one in homer counting only the containers in the right environments.

bastienwirtz (approved):
It's a small change, even if it's not common, no harm in that! Thanks.

Issue #347, opened 2022-01-07
Anonymous volume defined in Dockerfile blocks copying config.yml


scot-t:
**Issue**
The homer Dockerfile [specifies an anonymous volume][1], which prevents modifying any files under that node in subsequent steps [ref][2]

As far as I can tell (eg [this article][3]) this line could be safely removed and would make configuring homer containers simpler. Is this possible?

**Context**
I'm running homer in a container with docker compose, deployed to a remote host using docker contexts. This makes passing the configuration from my repo as a bind mount difficult. I would like to copy config to the image as a build step. Currently I just copy it to `/www/config.yml` and let [this workaround][4] take care of it.

[1]: https://github.com/bastienwirtz/homer/blob/main/Dockerfile#L32
[2]: https://docs.docker.com/engine/reference/builder/#notes-about-specifying-volumes
[3]: https://boxboat.com/2017/01/23/volumes-and-dockerfiles-dont-mix/
[4]: https://github.com/bastienwirtz/homer/blob/main/entrypoint.sh#L8

bastienwirtz:
Hi there, 

The Dockerfile does not include a volume instruction anymore :+1: 

Thanks,

Issue #209, opened 2021-04-18
Avoid adding example configs (and other files) after every start


RixzZ:
I have Homer with a docker volume mounted on '/www/assets' (which seems to be the recommended way to add custom configs).

After I added my config and a folder structure for the images I use, I deleted the rest of the example information that was on that directory. That would be:

- *dist* files
- *sample* files
- *tools* folder

Currently, after every container recreation from the image, that example information gets populated again to the directory, which I would like to avoid.

I see two possible solutions to this:

- Make that the information in that directory is added by the entrypoint script only if there isn't already information there (like when you have mounted a volume with your own config).
- Allow the user to configure an alternative config path (with an ENV variable maybe), so you could use that directory instead and avoid it being modified by the container.

Any thoughts on this?

crwallace2:
If you only want to store the config file externally so you can edit it, you could just mount the config.yml file to the container location for the config.yml, instead of mounting the whole folder. something like the below

`docker run -d \
  --name=homer \
  --restart unless-stopped \
  -p 8080:8080 \
  -e UID=1000 \
  -e GID=1000 \
  -v /vault/homer/config.yml/:/www/assets/config.yml \
  b4bz/homer:latest`

bastienwirtz:
Hi there ! 
Thanks for reporting this. The docker file has been reworked and now you can set the `INIT_ASSETS` env var to 0 to skip default asset install. Hope it helps!

Issue #193, opened 2021-02-22
[FR] Support docker --user switch


JuniorJPDJ:
Hi!
I'd like to run whole container as another user and I saw you are using user switch in entrypoint.
This is how redis added support for this: https://github.com/docker-library/redis/blob/master/6.0/docker-entrypoint.sh

bastienwirtz:
Hi there !
Thanks for reporting this. The docker file has been reworked to run as a non-root user, hope it works for you! feel free to give your feedback.

PR #421, opened 2022-04-10
Docker rework

bastienwirtz:
Hi there :wave: 

A lot issue has been raised around the docker container, so ... let's fix them all !

- https://github.com/bastienwirtz/homer/issues/325
- https://github.com/bastienwirtz/homer/issues/261
- https://github.com/bastienwirtz/homer/issues/209
- https://github.com/bastienwirtz/homer/issues/193

**I'm not a docker expert and I'm opening this PR to work with you to improve this, so feel free to share your thought, ideas, ... It will much appreciated.** If some of you are willing to try it, it would be awesome!

Homer is focused on simplicity, and really liked the idea of the container coming preconfigured with just a file to edit to get started. I tried, without success, to achieve that while not using the root user, so I decided to just simplify it and remove the entrypoint. I'm not even sure the `USER`, `GROUP`, `UID`, `GID` is usefull anymore ? **What to you think ?**

So, the only thing I can think of to keep it easy to get started is just to provide a better documentation to get started, or maybe a command to generate the initial asset ? 

This PR also bring a more robust http server (thanks @Kirmy34 !), and a simple way to use homer in a subpath. I kinda grouped all modifications related to docker / hosting as I want to avoid changing it too often (this kind of change may impact users)

While we're a it, I could use some advise on this one : https://github.com/bastienwirtz/homer/issues/347

Thanks a lot :heart:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,18 +12,19 @@ RUN yarn build
 # production stage
 FROM alpine:3.15
 
-ENV USER darkhttpd
-ENV GROUP darkhttpd
+ENV USER lighttpd
+ENV GROUP lighttpd
 ENV GID 911
 ENV UID 911
 ENV PORT 8080
 
 RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} && \
-    apk add -U --no-cache su-exec darkhttpd
+    apk add -U --no-cache lighttpd
 
 COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
 COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist/assets /www/default-assets
 COPY entrypoint.sh /entrypoint.sh
+COPY lighttpd.conf /lighttpd.conf
 
 HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
     CMD wget --no-verbose --tries=1 --spider http://127.0.0.1:${PORT}/ || exit 1
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,15 +17,16 @@ ENV GROUP lighttpd
 ENV GID 911
 ENV UID 911
 ENV PORT 8080
+ENV SUBFOLDER "/_"
 
 RUN addgroup -S ${GROUP} -g ${GID} && adduser -D -S -u ${UID} ${USER} ${GROUP} && \
     apk add -U --no-cache lighttpd
 
-COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
-COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist/assets /www/default-assets
 COPY entrypoint.sh /entrypoint.sh
 COPY lighttpd.conf /lighttpd.conf
 
+COPY --from=build-stage --chown=${USER}:${GROUP} /app/dist /www/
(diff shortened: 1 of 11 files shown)


bastienwirtz:
cc @ElleshaHackett, @git-noise,  @justsomebody42, @Nold360, @jorge-lavin, @werwolfby, @JuniorJPDJ.

bastienwirtz:
To give you an idea I tried some entrypoint like: 

```sh
#!/bin/sh

# Default assets & exemple configuration installation.
if [[ "${SKIP_ASSETS_INSTALL}" != "1" && ! -f "/www/config.yml" ]]; then
    echo "No configuration found, installing default config & assets"
    [[ ! -w "/www/assets/" ]] || echo "Assets directory not writable" # && exit 1

    while true; do echo n; done | cp -Ri /www/default-assets/* /www/assets/ &> /dev/null
    [[ $? -ne 0 ]] || echo "Fail to copy default assets" # && exit 1
    yes n | cp -i /www/default-assets/config.yml.dist /www/assets/config.yml &> /dev/null
    [[ $? -ne 0 ]] || echo "Fail to copy default config file" # && exit 1
fi

echo "Starting webserver"
lighttpd -D -f /lighttpd.conf
```

But as the user is created at build time, the assets directory will never have the necessary permission to be writable.

Nold360:
> But as the user is created at build time, the assets directory will never have the necessary permission to be writable.

I don't get that. User is created in Dockerfile, so he is in the image. The directory mounted as `/www/assets` should be owned by that user, before starting the container. [eg `fsGroup` in K8s][yea, afaik it's not that easy in plain docker.

Also on K8s we could use a init container to populate `/www/assets`. 

Anyways, why not simply check if there are any files in `/www/assets` and if not, copy the default ones? 


>  I'm not even sure the USER, GROUP, UID, GID is usefull anymore ? 
`UID` & `GID` don't have any real use afaik. We should have `USER` [in best case using UID eg. `USER 1000`] so Kubernetes can determine which user to give access to the PVCs.

bastienwirtz:
Thanks for your reply @Nold360. 

> I don't get that. User is created in Dockerfile, so he is in the image. The directory mounted as /www/assets should be owned by that user, before starting the container

I mean that no matter what you set as env vars, the homer user is defined at build time and will likely not have the right permissions to write into the mounted directory (unless the uid / guid match). 

> Anyways, why not simply check if there are any files in /www/assets and if not, copy the default ones? 

That is the current behavior. It works all the time because the container run as root, which will not longer be the case (the main goal of this PR). So unless the user inside the container (on which you have no control because it's created at build time) match the user on the host, you will have permission problems.

But I might be missing something.

Roundaround:
What all functionality is the web server meant to provide?

Is this something we could do with nginx instead of darkhttpd or lighttpd? I personally use nginx's docker container to deploy React apps all the time with extreme simplicity. This is one such example at its simplest:

```docker
# Build step
FROM node:16 as builder

WORKDIR /app

COPY . .

RUN yarn install
RUN yarn run build


# Serve step
FROM nginx:stable

WORKDIR /usr/share/nginx/html
RUN rm -rf ./*

COPY --from=builder /app/build .

ENTRYPOINT ["nginx", "-g", "daemon off;"]
```

bastienwirtz:
@Roundaround the web server is not the issue, it does not really matters. The point is how to install default assets (to keep the ease of use) without permission issues. But I think I know what to do, I'll push an update soon.

bastienwirtz:
All right, I think this is good to go.

PR #448, opened 2022-05-09
feat(pwa): enhance connectivity checks

### LINKED ISSUE #232: Authelia/Nginx redirects to download the /assets/config.yml file rather than display the page
I have Nginx with Authelia 2FA.  When I goto homer.TLD.com, I'm prompted for auth through Authelia.  After successful login, I get a download prompt for my yml config file at 
https://homer.TLD.com/authelia/?rd=https://homer.TLD.com/assets/config.yml

From what I understand, Authelia just redirects back to the requesting site which nginx is just proxying to `proxy_pass          http://<LAN IP>:80;`

Not sure what's causing this odd behavior.

bemble:
## Description

To check if user is offline, some simple tests without fetch request can be done (on `navigator.onLine`). But `navigator.onLine` does not mean you're online, some extra checks can be done in the right case.

I also check what happen to the `?alive` request, basically, if it was redirected, that means your behind an auth proxy which requires a re-auth.

Tested on Chrome/Brave/Safari for mac, Safari for iOS 15.

Fixes #232

I pushed a docker image, [`bemble/homer:fix-auth`](https://hub.docker.com/repository/docker/bemble/homer), for testing purpose, it will be deleted as soon as possible.

demo

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/src/components/ConnectivityChecker.vue
+++ b/src/components/ConnectivityChecker.vue
@@ -17,6 +17,9 @@ export default {
     };
   },
   created: function () {
+    if (/t=\d+/.test(window.location.href)) {
+      window.history.replaceState({}, document.title, window.location.pathname);
+    }
     let that = this;
     this.checkOffline();
 
@@ -61,7 +64,7 @@ export default {
         .then(function (response) {
           // opaqueredirect means request has been redirected, to auth provider probably
           if (response.type === "opaqueredirect" && !response.ok) {
-            window.location.reload(true);
+            window.location.href = window.location.href + "?t="+(new Date().valueOf());
           }
           that.offline = !response.ok;
         })
--- a/src/components/ConnectivityChecker.vue
+++ b/src/components/ConnectivityChecker.vue
@@ -56,7 +56,8 @@ export default {
 
       // extra check to make sure we're not offline
       let that = this;
-      return fetch(window.location.href + "?alive", {
+      const aliveCheckUrl = window.location.href + "?t="+(new Date().valueOf());
+      return fetch(aliveCheckUrl, {
         method: "HEAD",
         cache: "no-store",
         redirect: "manual"
@@ -64,7 +65,7 @@ export default {
         .then(function (response) {
           // opaqueredirect means request has been redirected, to auth provider probably
           if ((response.type === "opaqueredirect" && !response.ok) || [401, 403].indexOf(response.status) != -1) {
-            window.location.href = window.location.href + "?t="+(new Date().valueOf());
+            window.location.href = aliveCheckUrl;
           }
           that.offline = !response.ok;
         })
--- a/src/components/ConnectivityChecker.vue
+++ b/src/components/ConnectivityChecker.vue
@@ -29,15 +29,40 @@ export default {
       },
       false
     );
+    window.addEventListener(
+      "online",
+      function () {
+          that.checkOffline();
+      },
+      false
+    );
+    window.addEventListener(
+      "offline",
+      function () {
+        this.offline = true;
+      },
+      false
+    );
   },
   methods: {
     checkOffline: function () {
+      if (!navigator.onLine) {
+        this.offline = true;
+        return;
+      }
+
+      // extra check to make sure we're not offline
       let that = this;
       return fetch(window.location.href + "?alive", {
         method: "HEAD",
         cache: "no-store",
+        redirect: "manual"
       })
         .then(function (response) {
+          // opaqueredirect means request has been redirected, to auth provider probably
+          if (response.type === "opaqueredirect" && !response.ok) {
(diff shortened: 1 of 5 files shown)


fbartels:
With this change applied I am now getting the "you're offline friend" message when the auth token for my auth has expired (the alive check gets a 401 back), but there does not seem to be a way to login from that state.

Previously I just had the pwa disabled, which lead to less caching, but also a direct login when needed.

bemble:
> With this change applied I am now getting the "you're offline friend" message when the auth token for my auth has expired (the alive check gets a 401 back), but there does not seem to be a way to login from that state.
> 
> Previously I just had the pwa disabled, which lead to less caching, but also a direct login when needed.

Thanks for your reply. Could you give me more information about your configuration?

fbartels:
In a way its an easy setup, but also not one that is too easily replicated.

I am using a software called [Cloudron](https://cloudron.io) which needs to be installed on an Ubuntu box that is publicly reachable on port 80 and 443 (and does not allow any other software to be installed alongside of it). Cloudron has an app called [surfer](https://www.cloudron.io/store/io.cloudron.surfer.html) which is a static webserver with the ability to enforce authentication.

If you want to debug on an actual instance its probably the easiest if I setup a dedicated instance of surfer on my existing server for you.

In essence the current behaviour is that I get a 401 when the auth has expired (can be triggered by deleting a cookie) and only a force reload brings me back to the login page.

bemble:
> In a way its an easy setup, but also not one that is too easily replicated.
> 
> I am using a software called [Cloudron](https://cloudron.io) which needs to be installed on an Ubuntu box that is publicly reachable on port 80 and 443 (and does not allow any other software to be installed alongside of it). Cloudron has an app called [surfer](https://www.cloudron.io/store/io.cloudron.surfer.html) which is a static webserver with the ability to enforce authentication.
> 
> If you want to debug on an actual instance its probably the easiest if I setup a dedicated instance of surfer on my existing server for you.
> 
> In essence the current behaviour is that I get a 401 when the auth has expired (can be triggered by deleting a cookie) and only a force reload brings me back to the login page.

Thanks for the details, I pushed a new version, on [`bemble/homer:fix-auth`](https://hub.docker.com/repository/docker/bemble/homer), which _in theory_ handle your case. Thanks for testing!

fbartels:
Hi @bemble,

thanks for the additional change. I just made another local build with your latest code and can confirm that I am now getting the login page upon expiration of the session.

@bastienwirtz @Roundaround what do you think?

fbartels (approved):

src/components/ConnectivityChecker.vue
```diff
@@ -29,15 +32,40 @@ export default {
...
       let that = this;
       return fetch(window.location.href + "?alive", {
         method: "HEAD",
         cache: "no-store",
+        redirect: "manual"
       })
         .then(function (response) {
+          // opaqueredirect means request has been redirected, to auth provider probably
+          if ((response.type === "opaqueredirect" && !response.ok) || [401, 403].indexOf(response.status) != -1) {
+            window.location.href = window.location.href + "?t="+(new Date().valueOf());
```
bastienwirtz, line 67: I think we should use the same parameter everywhere, and the timestamp value as well, you should replace the ?alive parameter in the original request by the `t=<timestamp>`

bastienwirtz (changes requested):
Hi @bemble! 
Thanks for this ! I should have looked into it a long time ago. 
This implementation looks great, but I'm worried about potential race condition during the initial loading of the page with the main config loading request [here](https://github.com/bastienwirtz/homer/blob/main/src/App.vue#L233). 
Nothing guarantee the connectivity checker will run before it, and you might get an error first or not being redirected correctly ? Have you tried / experienced that? Anyway I think this [line](https://github.com/bastienwirtz/homer/blob/main/src/App.vue#L236) should be removed right ?

Last thing, could you add some documentation saying that if you want to use an auth proxy, you should activate the connectivity checker ?

Thanks a lot !

src/components/ConnectivityChecker.vue
bemble: I also think that if we can prevent the `alive` to be cached, we should do it.

bemble:
@bastienwirtz I'm diving a little more in the code and I saw that `connectivityCheck` is set to `true` in [`default.yml`](https://github.com/bastienwirtz/homer/blob/main/src/assets/defaults.yml#L11), it means that the test will be done at least once for users that don't rebuild homer with custom default config. The default behavior should always works, regardless user's value of `connectivityCheck`.
Anyway, code updated and documentation is done.

bastienwirtz (approved):
Looks good to me, thanks a lot @bemble !

PR #484, opened 2022-07-10
Add GET Support to Ping service

isuttell:
## Description

Adds an optional new property to the Ping custom service that lets the user leverage the GET method rather than the HEAD method. The HEAD method is not universally implemented which limits what the Ping service can actually monitor. 

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ X] I've read & comply with the [contributing guidelines](https://github.com/bastienwirtz/homer/blob/main/CONTRIBUTING.md)
- [ X] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. 
- [ X] I have made corresponding changes to the documentation (README.md).
- [ X] I've checked my modifications for any breaking changes, especially in the `config.yml` file

### DIFF
--- a/src/components/services/Ping.vue
+++ b/src/components/services/Ping.vue
@@ -29,7 +29,8 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      this.fetch("/", { method: "HEAD", cache: "no-cache" }, false)
+      const method = typeof this.item.method === 'string' && this.item.method.toLowerCase() === "get" ? "GET" : "HEAD"
+      this.fetch("/", { method, cache: "no-cache" }, false)
         .then(() => {
           this.status = "online";
         })
@@ -45,6 +46,8 @@ export default {
 .status {
   font-size: 0.8rem;
   color: var(--text-title);
+  white-space: nowrap;
+  margin-left: 0.25rem;
 
   &.online:before {
     background-color: #94e185;
--- a/src/components/services/Ping.vue
+++ b/src/components/services/Ping.vue
@@ -29,7 +29,13 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      const method = typeof this.item.method === 'string' && this.item.method.toLowerCase() === "get" ? "GET" : "HEAD"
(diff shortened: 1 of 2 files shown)


src/components/services/Ping.vue
```diff
@@ -45,6 +46,8 @@ export default {
 .status {
   font-size: 0.8rem;
   color: var(--text-title);
+  white-space: nowrap;
+  margin-left: 0.25rem;
```
bastienwirtz, line 50: Unrelated to the method support ? what does it changes ? If you could include a screenshot it would be awesome.

src/components/services/Ping.vue
```diff
@@ -29,7 +29,8 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      this.fetch("/", { method: "HEAD", cache: "no-cache" }, false)
+      const method = typeof this.item.method === 'string' && this.item.method.toLowerCase() === "get" ? "GET" : "HEAD"
```
bastienwirtz, line 32: I would prefer to check if the method is valid and stop if not. No unexpected behavior this way. Something like this: 
```js
const method = this.item.method.toUpperCase();
if (!['GET', 'HEAD', 'OPTION'].includes(method)) {
  console.error(`invalid ${method} ping method`);
  return;
}
```
Would be great.

bastienwirtz (changes requested):
Hi @isuttell,

Thanks for contributing! A few details but it look great :+1:

src/components/services/Ping.vue
isuttell: When the column is narrow the colored circle and text can appear on separate lines as seen at  https://imgur.com/a/rFmjJLe This change ensures that line break doesn't happen so the circle is always to the left of the text.

carlos-sarmiento:
Anything we can do to get this change landed?

bastienwirtz (approved):
Thanks for the changes, everthing looks good :+1:

PR #511, opened 2022-09-18
Service for rTorrent.

lindely:
## Description

I'm setting up homer for use as my browser start page and was adding links to various sites and (self-hosted) services. I thought it'd be interesting to include transfer rates and number of torrents in the link to my rTorrent web UI, so I made a service for that. I have also updated the `customservices.md` with instructions for the new service.

screenshot

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


src/components/services/Rtorrent.vue
```diff
@@ -0,0 +1,158 @@
...
+    // the specified method name and parsing the XML. The response
+    // is expected to adhere to the structure of a single numeric
+    // value.
+    getRate: async function(methodName, callback) {
+      return this.getXml(methodName)
+        .then((xml) => {
+          const rate = parseInt(xml.getElementsByTagName('value')[0].firstChild.textContent, 10);
+
+          if (callback) {
+            callback(rate);
```
bastienwirtz, line 104: You should directly `return rate` (so you would seemlessly get a promise with the value you need), this way, no need to mix callback & promise.

src/components/services/Rtorrent.vue
```diff
@@ -0,0 +1,158 @@
...
+          }
+
+          return Promise.resolve();
```
bastienwirtz, line 107: This is not necessary, by default the promise is returned.

src/components/services/Rtorrent.vue
```diff
@@ -0,0 +1,158 @@
...
+    this.fetchRates();
+    this.fetchCount();
+  },
+  methods: {
+    // Perform two calls to the XML-RPC service and fetch download
+    // and upload rates. Values are saved to the `ul` and `dl`
+    // properties.
+    fetchRates: async function() {
+      this.getRate('throttle.global_up.rate', (ul) => this.ul = ul)
+        .then(() => this.getRate('throttle.global_down.rate', (dl) => this.dl = dl))
```
bastienwirtz, line 86: it would be cool to use await instead of chaining promise, and use try/catch to handle errors.

src/components/services/Rtorrent.vue
```diff
@@ -0,0 +1,158 @@
...
+        });
+    },
+    // Fetch the numer of torrents by requesting the download list
+    // and counting the number of entries therein.
+    getCount: async function() {
+      return this.getXml('download_list')
+        .then((xml) => {
+          const arrayEl = xml.getElementsByTagName('array');
+          this.count = arrayEl ? arrayEl[0].getElementsByTagName('value').length : 0;
+          return Promise.resolve();
```
bastienwirtz, line 117: Here too you can remove the return.

bastienwirtz (changes requested):
Thanks for the beautifully made PR with all the comments & everything, it's very appreciated! :heart: . It's a cool service, thanks for adding it. 

A few notes here and there but it looks good :+1:

lindely:
@bastienwirtz I have committed a change which I hope will reflect the requested changes. I'm not a regular GitHub user and am unfamiliair with any proceedings regarding an update to my pull request, but since I see the new commit in the list I hope this notification message will suffice.

bastienwirtz:
Thanks for the update @lindely!
Could you save the api output as a static file and put it in the [`dummy-data`](https://github.com/bastienwirtz/homer/tree/main/dummy-data) directory so it can be tested without a rTorrent server (and also I will be able to add it to the demo page.
Thanks again !

src/components/services/Rtorrent.vue
```diff
@@ -0,0 +1,154 @@
...
+      }
+
+      return fetch(`${this.item.xmlrpc.replace(/\/$/, '')}/RPC2`, {
+        method: 'POST',
+        headers,
+        body: `<methodCall><methodName>${methodName}</methodName></methodCall>`
+      })
+      .then((response) => {
+        if (!response.ok) {
+          throw Error(response.statusText);
```
bastienwirtz, line 132: you can save you the error handling by using this.fetch ([custom fetch function](https://github.com/bastienwirtz/homer/blob/main/src/mixins/service.js#L15) that handle error detection. By default it does json parsing but you can disable it (last parameter to false).

src/components/services/Rtorrent.vue
```diff
@@ -0,0 +1,154 @@
...
+    // properties.
+    fetchRates: async function() {
+      try {
+        await this.getRate('throttle.global_up.rate')
+          .then((ul) => this.ul = ul);
+        await this.getRate('throttle.global_down.rate')
+          .then((dl) => this.dl = dl);
+      } catch {
+        this.error = true;
+      }
```
bastienwirtz, line 92: Sorry for not pointing that out in my first review but instead of waiting for the first request to finish to execute the second one (like your first implementation with the chain), why not run both request in parallel for faster execution? Something like (untested): 
```suggestion
       this.getRate('throttle.global_up.rate')
         .then((ul) => this.ul = ul)
         .catch(() => this.error = true); 
       this.getRate('throttle.global_down.rate')
         .then((dl) => this.dl = dl);
         .catch(() => this.error = true); 
```

bastienwirtz (changes requested):

lindely:
@bastienwirtz I added three dummy files for the three types of requests (download rate, upload rate and torrent list). All three lead to the same endpoint, **/RPC2** so I named the files after the API request names. I don't know how this will fit into your demo page but I hope this is usable.

The two awaits were removed. There was no specific reasoning behind not executing them in parallel.

Also I did not use the fetch method from the mixin since a user may have authentication set for their rTorrent instance, in which case the username and password have to be passed along in the HTTP header (now also added to `customservices.md` example) and the mixin doesn't allow for setting HTTP headers.

bastienwirtz (approved):
Thanks for the changes @lindely

PR #515, opened 2022-09-26
Update customservices.md

csleal:
"cicle" to circle


bastienwirtz (approved):
Hey @csleal , 
Good catch, thanks !

PR #523, opened 2022-10-10
Adds copy to clipboard custom service.

### LINKED ISSUE #500: Copy-to-Clipboard Option for Items
I think it would be awesome if you could copy predefined text to the clipboard instead of opening an URL. This could be useful especially for applications like Teamspeak Servers or Game Servers.

I would suggest adding a config key to the `Item` called something like `clipboard`. When defined, the text specified will be copied to the clipboard when clicking the item and a notification would pop up to indicate that the text has been copied. Also, there could be a little indicator on the item to show that the URL will be copied on clicking the item.

My current workaround "abuses" the `javascript` URL method, but I think it would be awesome if this would be supported natively.

```yml
- name: "Game Server"
    icon: "fas fa-gamepad"
    items:
      - name: "mc.zekro.de"
        subtitle: "Public Vanilla Minecraft Server"
        logo: "assets/tools/minecraft.png"
        keywords: "minecraft game ichigo"
        url: "javascript:navigator.clipboard.writeText('mc.zekro.de')&&alert('Address has been copied to the clipboard.')"
```

luixal:
## Description

Adds a custom service that adds an icon button on the indicator to copy to the clipboard the text in the `clipboard` field.

Fixes #500 

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/CopyToClipboard.vue
+++ b/src/components/services/CopyToClipboard.vue
@@ -1,7 +1,7 @@
 <template>
   <Generic :item="item">
     <template #indicator>
-      <div class="status"><i class="fa-regular fa-copy fa-xl" :class="{'fa-beat': animate, 'copied-icon': animate}" @click="copy"></i></div>
+      <div class="status"><i ref="copyIcon" class="fa-regular fa-copy fa-xl" :class="{'scale': animate}" @click="copy($event)"></i></div>
     </template>
   </Generic>
 </template>
@@ -22,24 +22,62 @@ export default {
   data: () => ({
     animate: false
   }),
+  mounted: function() {
+    this.$refs.copyIcon.addEventListener(
+      'animationend',
+      (event) => { this.animate = false; });
+  },
   methods: {
-    copy() {
+    copy(event) {
       javascript:navigator.clipboard.writeText(this.item.clipboard)
       this.animate = true;
-      setTimeout(
-        () => { this.animate = false },
-        1000
-      );
     }
   },
 };
 </script>
 
 <style scoped lang="scss">
-.status {
+.scale {
+	-webkit-animation: scale-up 0.3s cubic-bezier(0.250, 0.460, 0.450, 0.940) both;
+	        animation: scale-up 0.3s cubic-bezier(0.250, 0.460, 0.450, 0.940) both;
+}
+.is-light i {
   color: black;
 }
-.copied-icon {
-  color: #5dbc9d;
+.is-dark i {
+  color: white;
+}
+  /**
+ * ----------------------------------------
+ * animation scale-down-center
+ * ----------------------------------------
+ */
+@-webkit-keyframes scale-up {
+  0% {
+    -webkit-transform: scale(1);
+            transform: scale(1);
+  }
+  50% {
+    -webkit-transform: scale(1.25);
+            transform: scale(1.25);
+  }
+  100% {
+    -webkit-transform: scale(1);
+            transform: scale(1);
+  }
+}
+@keyframes scale-up {
+  0% {
+    -webkit-transform: scale(1);
+            transform: scale(1);
+  }
+  50% {
+    -webkit-transform: scale(1.25);
+            transform: scale(1.25);
+  }
+  100% {
+    -webkit-transform: scale(1);
+            transform: scale(1);
+  }
 }
 </style>
--- a/src/components/services/CopyToClipboard.vue
+++ b/src/components/services/CopyToClipboard.vue
@@ -1,0 +1,45 @@
+<template>
+  <Generic :item="item">
+    <template #indicator>
+      <div class="status"><i class="fa-regular fa-copy fa-xl" :class="{'fa-beat': animate, 'copied-icon': animate}" @click="copy"></i></div>
+    </template>
+  </Generic>
+</template>
+
+<script>
+import service from "@/mixins/service.js";
+import Generic from "./Generic.vue";
+
+export default {
+  name: "CopyToClipboard",
+  mixins: [service],
+  props: {
+    item: Object,
+  },
+  components: {
+    Generic,
+  },
+  data: () => ({
+    animate: false
+  }),
+  methods: {
+    copy() {
+      javascript:navigator.clipboard.writeText(this.item.clipboard)
+      this.animate = true;
+      setTimeout(
+        () => { this.animate = false },
+        1000
+      );
+    }
+  },
+};
+</script>
+
+<style scoped lang="scss">
+.status {
+  color: black;
+}
+.copied-icon {
+  color: #5dbc9d;
+}
+</style>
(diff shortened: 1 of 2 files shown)


luixal:
I guess this can also be made easily on the generic component. I can change it to that kind of implementation, but maybe it's better to keep things splitted apart.

Consider adding the hacktoberfest labels! :)

src/components/services/CopyToClipboard.vue
```diff
@@ -0,0 +1,45 @@
...
+    Generic,
+  },
+  data: () => ({
+    animate: false
+  }),
+  methods: {
+    copy() {
+      javascript:navigator.clipboard.writeText(this.item.clipboard)
+      this.animate = true;
+      setTimeout(
```
bastienwirtz, line 29: I think it shoud be possible to avoid the setTimeout by using CSS [`animation-delay`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay), that would be great.

src/components/services/CopyToClipboard.vue
```diff
@@ -0,0 +1,45 @@
...
+        () => { this.animate = false },
+        1000
```
bastienwirtz, line 31: I not against the animation but I think it should be very fast, to it's discreet & subtle, you know what I mean ?

src/components/services/CopyToClipboard.vue
```diff
@@ -0,0 +1,45 @@
...
+      );
+    }
+  },
+};
+</script>
+
+<style scoped lang="scss">
+.status {
+  color: black;
```
bastienwirtz, line 40: An alternative color for dark mode would be great!

bastienwirtz: Also, maybe a litlle change of color on hover ? (has the entire card is clickable, it would confirm that you are about to activate on the action button)

bastienwirtz (changes requested):
Thanks for the implementation @luixal. I think we can keep this as a separate service.
A few minor things, aside from that it looks good.

src/components/services/CopyToClipboard.vue
luixal: I implemented it this way to use the font-awesome animation class. I can  remove this by including a custom animation in CSS. It's related to the other change about making the animation shorter.

src/components/services/CopyToClipboard.vue
luixal: As said before, I didn't control animation speed to avoid including a custom CSS animation (which I think it's better, just a fast "blink" or pulse) just because I though you would prefer it that way.

I agree to your comments so I'll include the custom animation in CSS :)

src/components/services/CopyToClipboard.vue
luixal: Yep, lessons learned from the proxmox service changes. I will look into bulma modifier classes because I'm not sure how to include a CSS selector that takes in count the `is-dark` class outside the component (remember, I'm not really good at frontend stuff...).

luixal:
Updated. As the animation is fast now, I removed the color alternation to green. Also, removed the timeout, using `animationend` event listener instead. Also included styles for light and dark themes.

Looks like this now:

chrome-capture-2022-9-17

this for dark mode:

chrome-capture-2022-9-17 (1)

what do you think?

src/components/services/CopyToClipboard.vue
```diff
@@ -0,0 +1,83 @@
...
+    animate: false
+  }),
+  mounted: function() {
+    this.$refs.copyIcon.addEventListener(
+      'animationend',
+      () => { this.animate = false; });
+  },
+  methods: {
+    copy() {
+      javascript:navigator.clipboard.writeText(this.item.clipboard)
```
bastienwirtz, line 32: Is there a reason to use the `javascript:` prefix? I think this should be fine:  

```suggestion
      navigator.clipboard.writeText(this.item.clipboard)
```

src/components/services/CopyToClipboard.vue
```diff
@@ -0,0 +1,83 @@
...
+  },
+  components: {
+    Generic,
+  },
+  data: () => ({
+    animate: false
+  }),
+  mounted: function() {
+    this.$refs.copyIcon.addEventListener(
+      'animationend',
```
bastienwirtz, line 27: If you use `@animationend`, you should be able to remove this block

src/components/services/CopyToClipboard.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <Generic :item="item">
+    <template #indicator>
+      <div class="status"><i ref="copyIcon" class="fa-regular fa-copy fa-xl" :class="{'scale': animate}" @click="copy()"></i></div>
```
bastienwirtz, line 4: I believe you can use `@animationend` instead of the `ref` + `mounted` function. 
```suggestion
      <div class="status"><i @animationend="animate=false" class="fa-regular fa-copy fa-xl" :class="{'scale': animate}" @click="copy()"></i></div>

```

bastienwirtz (changes requested):

src/components/services/CopyToClipboard.vue
luixal: True, there's no need to use the `javascript:` prefix. I just took the code from the example in the issue. Removed.

src/components/services/CopyToClipboard.vue
luixal: True, no need for setting the listener manually.

src/components/services/CopyToClipboard.vue
luixal: True. Removed.

bastienwirtz (commented):
Looks good to me :+1:

luixal:
Hi @bastienwirtz , were (or are) you able to merge this? As I accidentally removed my previous fork, I'm not sure how can this be handled.

luixal:
Will do that. I've also made a mess of the new fork, in fact, two of my PRs look just like the same (I swear I didn't mean to replicate it) so, once you've merged one of them (or discard, as you wish), close the other and I will clean all stuff up.

Meanwhile, I'll make another fork an recover branches with the patches, not sure if it will take the same PR or a different one.

Sorry for this mess...

luixal:
Well... it seems that if I put the same to the branch, github takes it at the same PR.

Changes done! :)

bastienwirtz:
@luixal hum I'm still unable to merge this request :cry: 
You may have to close and reopen the pull request. 

(If you want I can apply the patch myself but it won't appear as your contribution)

luixal:
I couldn't make the patch thing work, I copy&paste changes to my branch.

Feel free to apply the patch. Same for the other PR. I made a total mess with this xD

bastienwirtz:
@luixal Patch added to the main branch :+1:

PR #533, opened 2022-10-17
Proxmox improvements 2

### LINKED ISSUE #531: Proxmox card cutting off info
Per @luixal I am submitting this issue. 
On small screens the data is cut off. Here's an example: https://i.postimg.cc/fy0CBv4z/Screenshot-2022-10-14-at-13-19-43-207-LAN-net-Dashboard-Intranet.png

luixal:
## Description

Extends Proxmox service:

- Adds more configuration variables.
- Adds LXC containers support.
- Some minor tweaks and refactor.

Fixes #531 

## Type of change

## Checklist:

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -272,7 +272,9 @@ Configuration example:
   warning_value: 50
   danger_value: 80
   api_token: "PVEAPIToken=root@pam!your-api-token-name=your-api-token-key"
-  small_font_on_small_screens: true # uses a smaller font for subtitle on small screens
+  # values below this line are optional (default value are false/empty):
   hide_decimals: true # removes decimals from stats values.
-  hide: ["vms", "lxc", "disk", "mem", "cpu"] # hides values included in the array
+  hide: ["vms", "vms_total", "lxcs", "lxcs_total", "disk", "mem", "cpu"] # hides values included in the array
+  small_font_on_small_screens: true # uses small font on small screens (like mobile)
+  small_font_on_desktop: true # uses small font on desktops (just in case you're showing much info)
 ```
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -272,6 +272,7 @@ Configuration example:
   warning_value: 50
   danger_value: 80
   api_token: "PVEAPIToken=root@pam!your-api-token-name=your-api-token-key"
-  small_font_on_small_screens: true #uses a smaller font for subtitle on small screens
+  small_font_on_small_screens: true # uses a smaller font for subtitle on small screens
   hide_decimals: true # removes decimals from stats values.
+  hide: ["vms", "lxc", "disk", "mem", "cpu"] # hides values included in the array
 ```
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -272,5 +272,6 @@ Configuration example:
   warning_value: 50
   danger_value: 80
   api_token: "PVEAPIToken=root@pam!your-api-token-name=your-api-token-key"
-  small_font_on_small_screens: true #optional: uses a smaller font for subtitle on small screens
+  small_font_on_small_screens: true #uses a smaller font for subtitle on small screens
+  hide_decimals: true # removes decimals from stats values.
 ```
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -272,4 +272,5 @@ Configuration example:
   warning_value: 50
   danger_value: 80
   api_token: "PVEAPIToken=root@pam!your-api-token-name=your-api-token-key"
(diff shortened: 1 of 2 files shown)


docs/customservices.md
```diff
@@ -272,4 +272,9 @@ Configuration example:
   warning_value: 50
   danger_value: 80
   api_token: "PVEAPIToken=root@pam!your-api-token-name=your-api-token-key"
+  # values below this line are optional (default value are false/empty):
+  hide_decimals: true # removes decimals from stats values.
+  hide: ["vms", "vms_total", "lxcs", "lxcs_total", "disk", "mem", "cpu"] # hides values included in the array
+  small_font_on_small_screens: true # uses small font on small screens (like mobile)
```
bastienwirtz, line 278: I think this one should be automatic using CSS & [bulma responsive classes](https://bulma.io/documentation/overview/responsiveness/#breakpoints).

docs/customservices.md
```diff
@@ -272,4 +272,9 @@ Configuration example:
   warning_value: 50
   danger_value: 80
   api_token: "PVEAPIToken=root@pam!your-api-token-name=your-api-token-key"
+  # values below this line are optional (default value are false/empty):
+  hide_decimals: true # removes decimals from stats values.
+  hide: ["vms", "vms_total", "lxcs", "lxcs_total", "disk", "mem", "cpu"] # hides values included in the array
```
bastienwirtz, line 277: I think it could be better to me to choose what to show, what do you think ?

docs/customservices.md
```diff
@@ -272,4 +272,9 @@ Configuration example:
...
+  small_font_on_small_screens: true # uses small font on small screens (like mobile)
+  small_font_on_desktop: true # uses small font on desktops (just in case you're showing much info)
```
bastienwirtz, line 279: Not sure about this one either, it could be automatic based on the number of items choosen? :thinking:

bastienwirtz (changes requested):
Thanks for this update @luixal, aside from a few things, it looks good :+1:

docs/customservices.md
luixal: I think most users want to show as much info as possible, that's why I thought it was better to put what you want to hide. Following that line, you don't have to add an extra config line if you want to show all info.

Anyhow, I can change the implementation if your think it's better to use show instead of hide. In that case, I would change everything as I don't like to mix show and hide fields. Your choice :)

docs/customservices.md
luixal: I think most users want to show as much info as possible, that's why I thought it was better to put what you want to hide. Following that line, you don't have to add an extra config line if you want to show all info.

Anyhow, I can change the implementation if your think it's better to use show instead of hide. In that case, I would change everything as I don't like to mix show and hide fields. Your choice :)

docs/customservices.md
luixal: Same thing, I was thinking about big screen sizes (ultra wide monitors). Right now, max wide is a fixed value, but I was willing to try and make a PR allowing that to be changed :)

If you feel like you're gonna keep the max wide fixed, I can make this one automatic according to which fields are shown.

luixal: I've been playing around to see the edge cases of setting font size automatically **only on desktop** and there are several things involved:

- If I `hide_decimals` and `*_total`, the font is good at the default size.
- If I don't `hide_decimals` but hide `*_total` small font is needed.
- If I hide any other field, default font size is great.

I can write this logic to make this automatic, is not too hard as I build and array with fields to hide (as vms and lxcs are hidden by default if none of them exist in proxmox), but not sure if that's the best way to go.

docs/customservices.md
luixal: Sorry, I wrote this for desktop version frist (see below on your other comment) but thing is easier here. I can set smaller font automatically if a field which is not a total is hidden.

Anyhow, as there are so many screen sizes on mobile... I can set it by default and let and `avoid` field in case someone needs it.

Do you feel better about that option?  I can apply the same to desktop.

luixal:
Pfff... I made a mistake.

I just deleted my fork thinking I was removing a branch. Yes I didn't read the warning message :(

I can clone this PR, but I'm not sure if I can push changes to the same PR if I make a new branch with the same name. Just leave your comment and will try to make my best :)

docs/customservices.md
luixal: I will recover this PR as the old one and merge the main branch to resolve conflicts. Do you wanto to go deeper into this or should we resolve this conversation?

If in the future you add a custom class to be used, I can update this to the newer styles.

bastienwirtz:
@luixal Patch added to the main branch :+1:

PR #534, opened 2022-10-22
Fixed the table in the Proxmox section

tralce:
Just a simple documentation improvement.


bastienwirtz (approved):
Good catch @tralce, thanks for contributing!

Issue #468, opened 2022-06-16
Homer shows "You're Offline Friend" after 30 seconds


rarick:
# My setup
- traefik as a reverse proxy & domain controller
- [thomseddon/traefik-forward-auth](https://github.com/thomseddon/traefik-forward-auth) as my forward auth provider
- auth0 as my oauth provider
  - All my domains and wildcards are entered in auth0 for CORS/origins

# The issue
Homer is behind the authorization proxies. I am seeing "You're offline friend" after 30 seconds on homer. I'm not offline, and I can always reload the page without issue (hard reload as well). I see that the request is to `mydomain.com?t=<unix timestamp>`. This is receiving a 500 error, but I don't know where this response is coming from.

I see the request callstack starts with `checkOffline`.

Any ideas why this is happening?

bastienwirtz:
Hi @rarick

Hum, that's weird. The connectivity check request (which can be disabled) if an `HEAD` request, maybe traefik block this kind of request? Can you tell if the request hit the container or not?
Is there a message associated with the 500 error ? Can you also check the container logs?

zhenry9:
I have a nearly identical issue. 

Setup:

 - traefik2
 - thomseddon/traefik-forward-auth
 - Google oauth

I also get the "offline friend" message after 30 seconds, which is solved by refreshing. You mentioned the connectivity check could be disabled. How do you do this?

Fratt:
Same problem here

vincentvriends:
I have the same issue, but instead of traefik I use the Explicit Proxy build into a FortiGate firewall.
When I check the developer tools in my browser I see that a request is made to the following URL: http://dashboard.int.mydomain.be//index.html?t=1660243712402

So with 2 // instead of 1, this request gives back a 504 gateway timeout.

bastienwirtz:
Thanks for the feedback, @zhenry9. Disabling the connectivity checker will workaround this issue.
@rarick @Fratt @vincentvriends Have you tried that?

That being said, the connectivity checker should work, it just periodically send a request to the server and execpt either an http response code in the  200-299 range OR a redirection (to handle auth proxy redirection, il this case the whole page reload and follow the redirection). If you don't need that, you should disable it anyway.

c0pper:
Hi, where exactly do you find this connectivity check option? my config.yml only has things that should or hould not appear on the html page and theme related stuff... Thanks

PR #535, opened 2022-10-24
ConnectivityChecker: Remove double // in aliveCheckUrl causing Invalid Redirect URL with CloudFlare Access.

bchavez:
### Bug Fix
When Homer is protected behind **CloudFlare Access**, the `ConnectivityChecker` builds an `aliveCheckUrl` with something like:

```
https://homer.domain.net//index.html?t=1666582254399
```
and makes an HTTP `HEAD` to the `homer.domain.net` to verify connectedness.

The problem arises when the cookie protecting `homer.domain.net` is expired; and results in an `HTTP 302` redirect to a login page.

The browser follows the `HTTP 302` redirect and the user lands on a site like:

```
https://__team__.cloudflareaccess.com/cdn-cgi/access/login/....redirect_url=%2F%2Findex.html%3Ft%3D1666582258096....
```

Notice the URL encode `redirect_url=%2F%2Findex.html`. Decoded, the `redirect_url` looks something like `//index.html`. The double `//` in the `redirect_url` causes **CloudFlare Access** to display the following error page (instead of a proper Login Page or Redirect):

image

If we remove the leading `%2F` from the `redirect_url` so that we have: `redirect_url=%2Findex.html...`; we get the expected login page and/or a proper redirect with the necessary authorization cookies.

Feel free to push modifications or changes to this PR.

Thanks,
Brian

## Type of change

## Checklist:

### DIFF
--- a/src/components/ConnectivityChecker.vue
+++ b/src/components/ConnectivityChecker.vue
@@ -56,9 +56,8 @@ export default {
 
       // extra check to make sure we're not offline
       let that = this;
-      const aliveCheckUrl = `${window.location.origin}${
-        window.location.pathname
-      }/index.html?t=${new Date().valueOf()}`;
+      const urlPath = window.location.pathname.replace(/\/+$/, "");
+      const aliveCheckUrl = `${window.location.origin}${urlPath}/index.html?t=${new Date().valueOf()}`;
       return fetch(aliveCheckUrl, {
         method: "HEAD",
         cache: "no-store",


bastienwirtz (commented):
Hi @bchavez, 

Indeed, the `location.pathname` can contain a trailing slash and cause a double slash. Thanks for this fix!

bastienwirtz (approved):

XxAcielxX:
Same issue is happening if you use Authelia as an authentication method. After the login, it redirects to `https://mydomain.ltd//index.html?t=1667722753069`

XxAcielxX:
I'm having a side affect of this fix with the redirecting. Now if I'm on `https://mydomain.ltd//index.html`, the redirect adds another
`/index.html?t=xxxxxxxxxxx` and I end up with `https://mydomain.ltd/index.html/index.html?t=xxxxxxxxxxx`

martin-micimo:
Yea. Same as #468

Why was the index.html hardcoded here? Works fine if I remove "/index.html" from aliveCheckUrl in the compiled js.

Issue #536, opened 2022-10-24
"Add to homescreen" on Android points to the wrong address


agilly1989:
After navigating to the correct address (in my case `https://home.gil.ly`) the page appears correctly in Chrome on my phone. I then "Add to Homescreen" and after it goes through the motions, I tap on the icon it made and it comes up with a phishing warning.

After some investigating, it seems to be making a shortcut to `https://home.gil.ly/assets/` even though I am on `https://home.gil.ly`

And if you are curious, I am running self-signed certs and I have installed them onto my phone so that is not the issue.

(note, this domain is a my network only domain via some fun dnsmasq stuff on my router)

BonexP:
same to me.
i download release and host it using caddy.it works.
but when i install it.it doesnt work.

fbartels:
I can reproduce this on a new system as well.

bastienwirtz:
Hi there, thanks for reporting this, [the fix](https://github.com/bastienwirtz/homer/pull/548) has been merged, it should be better now :+1:

agilly1989:
Upgraded to docker image [v22.11.1](https://hub.docker.com/layers/b4bz/homer/v22.11.1/images/sha256-4db1f87fe09890a430451994b7fb9def7f0b6e7c305c8dbe26640bc4f7ba6c95?context=explore) and it seems to install as an app instead of a web shortcut now (Chrome on Android and on PC) but is still pointing to ./assets/

![image](https://user-images.githubusercontent.com/50060406/201323491-1eabafcc-b21a-4a11-a37a-385e129c7e78.png)

nikkopt:
I just tried this and am also getting a 403 forbidden error. Works fine in the browser.

nikkopt:
Proof:
![1](https://github.com/bastienwirtz/homer/assets/2060960/82030c64-5410-44ce-bee6-8dc9fd2c5bf2)
![Screenshot_20230906-203614](https://github.com/bastienwirtz/homer/assets/2060960/9ff66671-4d9d-4e03-b08d-225d4e3fbfe2)

Using latest docker image as of this day.

PR #548, opened 2022-11-07
Add scope and start_url to PWA manifest

### LINKED ISSUE #544: 403 Forbidden error when opening from Android homepage
Hello, I'm getting a 403 Forbidden error when opening my Homer dashboard from a shortcut on the Android launcher in both Brave Browser and Google Chrome.  Both browsers can access the URL (http://192.168.0.20:8078/) with no issues when accessing the address from within the browser.  However, when I Add to Home Page the launcher shortcut doesn't work and a 403 Forbidden error is displayed.

I have another dashboard, Heimdall, which works with no issues as above and I haven't encountered this error with any other application on my home server.  I'm not sure why only the launcher shortcut doesn't work, can you help?

wipash:
## Description

Without explicitly defining `start_url` and `scope` URLs in the PWA manifest, Vite defaults them to `./` when it generates the manifest. Because the manifest path is within the `assets` folder, this essentially results in the installed desktop app on mobile devices pointing to `https://homer.yourdomain/assets/`.

The `start_url` and `scope` URLs in the PWA manifest are relative, so given that manifest.json lives in assets/, these should point to the root of the site. Alternatively they could be set to `/` instead of `../`, but I felt that pointing to the parent directory would cover more possible implementations.

Fixes #544 and #536 

## Type of change

## Checklist:

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -21,6 +21,8 @@ export default defineConfig({
         short_name: "Homer",
         description: "Home Server Dashboard",
         theme_color: "#3367D6",
+        start_url: "../",
+        scope: "../",
         icons: [
           {
             src: "./icons/pwa-192x192.png",


bastienwirtz (approved):
Indeed, that should be better :+1: 
Thanks a lot @wipash!

PR #580, opened 2022-12-27
Implement PiHole API token

### LINKED ISSUE #579: PiHole stats not available in Homer if WebUI is password protected
### Describe the bug

Everything is functional, except PiHole stats are not displayed. There are no CORS related errors, I fixed the issue into #478.

### Expected behaviour

PiHole stats show in Homer.

### Logs & errors

PiHole runs on a dedicated Raspberry Pi and stats show, when called directly:

image

Browser console shows no errors:

image

PiHole.abc61540.js Preview

image

PiHole.abc61540.js Headers

image

### Configuration

image

```yml
services:
  - name: "Network"
    icon: "fas fa-server"
    items:
      - name: "chaos.lan"
        type: "PiHole"
        logo: "assets/tools/pihole.png"
        tag: "dns"
        keywords: "dns pihole"
        url: "http://chaos.lan/admin/"
```

fmunteanu:
## Description

When PiHole web interface is password protected, a query string containing an API token is required for validation.

Fixes #579

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -38,7 +38,7 @@ export default {
   }),
   computed: {
     apiQuery() {
-      return this.item.apikey ? "?summaryRaw&auth=${this.item.apikey}" : "";
+      return this.item.apikey ? "`?summaryRaw&auth=${this.item.apikey}`" : "";
     },
     percentage: function () {
       if (this.ads_percentage_today) {
@@ -52,7 +52,7 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      const result = await this.fetch("/api.php${this.apiQuery}")
+      const result = await this.fetch("`/api.php${this.apiQuery}`")
         .catch((e) => console.log(e));
 
       this.status = result.status;
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -37,6 +37,9 @@ export default {
     ads_percentage_today: 0,
   }),
   computed: {
+    apiQuery() {
+      return this.item.apikey ? "?summaryRaw&auth=${this.item.apikey}" : "";
+    },
     percentage: function () {
       if (this.ads_percentage_today) {
         return this.ads_percentage_today.toFixed(1);
@@ -49,7 +52,8 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      const result = await this.fetch("/api.php").catch((e) => console.log(e));
+      const result = await this.fetch("/api.php${this.apiQuery}")
+        .catch((e) => console.log(e));
 
       this.status = result.status;
       this.ads_percentage_today = result.ads_percentage_today;
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -37,9 +37,6 @@ export default {
     ads_percentage_today: 0,
   }),
   computed: {
-    apiQuery() {
-      return this.item.apikey ? "`?summaryRaw&auth=${this.item.apikey}`" : "";
-    },
     percentage: function () {
       if (this.ads_percentage_today) {
         return this.ads_percentage_today.toFixed(1);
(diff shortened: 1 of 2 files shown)


src/components/services/PiHole.vue
```diff
@@ -49,7 +52,8 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      const result = await this.fetch("/api.php").catch((e) => console.log(e));
+      const result = await this.fetch("/api.php${this.apiQuery}")
```
fmunteanu, line 55: Might require backticks for `${this.apiQuery}`.

tibbe: I believe it should be:

```js
const result = await this.fetch(`/api.php${this.apiQuery()}`)
```

src/components/services/PiHole.vue
```diff
@@ -37,6 +37,9 @@ export default {
     ads_percentage_today: 0,
   }),
   computed: {
+    apiQuery() {
+      return this.item.apikey ? "?summaryRaw&auth=${this.item.apikey}" : "";
```
tibbe, line 41: I believe this needs to be:

```js
return this.item.apikey ? `?summaryRaw&auth=${this.item.apikey}` : "";
```

fmunteanu: @tibbe I was looking at other files and @bastienwirtz is usually adding backticks to the function call, something like:

```
const result = await this.fetch(`/api.php${this.apiQuery()}`)
```

tibbe: I believe you need the backticks every time you do variable interpolation (i.e. every time you mention `${..}`). which happens on this line. Perhaps it's possible to delay the interpolation but that seems error prone (e.g. the interpolated expression might not be in scope at the later point).

fmunteanu: @tibbe thank you for the info, not experienced with due language, Python head here. :) Basically, I should add the backticks in both locations, right? If yes, I'll push the changes now, let me know please.

tibbe: Yes I believe they need to be in both places and contain the entire string (like I did I my examples).

fmunteanu: @tibbe, I pushed the change.

src/components/services/PiHole.vue
```diff
@@ -37,6 +37,9 @@ export default {
     ads_percentage_today: 0,
   }),
   computed: {
+    apiQuery() {
```
bastienwirtz, line 40: Computed property are used to get reactivity in property used in the UI, we don't need that here as this is not displayed. As it's just a simple condition, you can add it to the fetchStatus directly.

src/components/services/PiHole.vue
```diff
@@ -37,6 +37,9 @@ export default {
...
+      return this.item.apikey ? "`?summaryRaw&auth=${this.item.apikey}`" : "";
```
bastienwirtz, line 41:
```suggestion
      return this.item.apikey ? `?summaryRaw&auth=${this.item.apikey}` : "";
```
backticks are the same as doublequote + string interpolation. it replace the doublequote.

src/components/services/PiHole.vue
```diff
@@ -49,7 +52,8 @@ export default {
   },
   methods: {
     fetchStatus: async function () {
-      const result = await this.fetch("/api.php").catch((e) => console.log(e));
+      const result = await this.fetch("`/api.php${this.apiQuery}`")
```
bastienwirtz, line 55: Something like this should to the job
```suggestion
      const authQueryParams = this.item.apikey ? `?summaryRaw&auth=${this.item.apikey}` : "";
      const result = await this.fetch(`/api.php${authQueryParams}`)
```

bastienwirtz (changes requested):
Just saw your open issue and this PR. Thanks @fmunteanu !
Looks good, a few little thing and we should be good to go.

bastienwirtz:
Thanks for the adjustment @fmunteanu I will test it asap.

bastienwirtz (approved):

bastienwirtz:
Thanks again @fmunteanu :+1:

ihardyslide:
Hi,

this fix was working fine until yesterday, when I updated to the latest version of Pi-Hole (Pi-hole v5.15.5, FTL v5.21, Web Interface v5.18.4).
Since then I experience this issue again.
My analysis shows that the API call just returns empty array, so Homer does not show any addtional info:
image

I doublechecked the config and the API key is still correct.

Anybody else having this issue again?

Thx & Br

Issue #597, opened 2023-01-30
Option to bundle common dashboard icons


wyvernzora:
**Is your feature request related to a problem? Please describe.**
When building out dashboard, I need to separately mount a volume with service icon images. It would be nice to have an option to bundle common service icons into the container.

**Describe the solution you'd like**
I propose creating a separate set of docker tags that include bundled icon images. A good source of them can be https://github.com/walkxcode/dashboard-icons

**Describe alternatives you've considered**
I am currently extending the homer image with my own Dockerfile like the following
```
FROM bitnami/git AS builder
WORKDIR /root
RUN git clone https://github.com/walkxcode/dashboard-icons.git

FROM b4bz/homer
COPY --from=builder /root/dashboard-icons/png /www/assets/dashboard-icons/png
COPY --from=builder /root/dashboard-icons/svg /www/assets/dashboard-icons/svg
```

However, given that homer is a dashboard, and most people running it would want to have nice icons, bundling it into homer itself can be useful. Downside is that it increases the image size, therefore I propose a new set of image tags instead of simply bundling into all of the homer releases.

FriendlyNGeeks:
thats pretty strong. I thought my forked repo was pretty good for the icons that I upload to homer each time. [https://github.com/FriendlyNGeeks/homer-icons](https://github.com/FriendlyNGeeks/homer-icons) ofc I added some custom ones for my environment. Def agree with OP for a included bundle.

PR #598, opened 2023-01-31
Bundle dashboard-icons into the image

wyvernzora:
## Description
This change bundles the https://github.com/walkxcode/dashboard-icons into the container image so that out of the box it provides icon images to most of the common services one would host in a home lab environment. The downside of this change is added image size and an extra dependency.

If bundling into the main image is not desirable, we can also introduce an alternative set of tags that incorporate the icons. There are already precedents of dashboards integrating with the dashboard-icons project, as described in the dashboard-icons README.

Discussion issue #597 

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,11 @@ RUN yarn install --frozen-lockfile
 COPY . .
 RUN yarn build
 
+# dashboard icons stage
+FROM bitnami/git as icons-stage
+WORKDIR /root
+RUN git clone https://github.com/walkxcode/dashboard-icons.git
+
 # production stage
 FROM alpine:3.16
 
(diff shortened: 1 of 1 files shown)


bastienwirtz:
Hello, 

Thanks for your proposition @wyvernzora!
As you mention, it wouldn't be great to add it to the main image, and I'm not sure building another image worth it (also, it opens a path for infinite amount of images to build, if we do this for @walkxcode why not other?). 

As an alternative, I added a link to @walkxcode repo in the demo config as well as in the [tips & tricks](https://github.com/bastienwirtz/homer/blob/main/docs/tips-and-tricks.md#dashboard-icons) section, so user will be able to find the icons easily. 

@wyvernzora if you want you could also update the doc to explain how to build your own homer image with the icons included ?

PR #603, opened 2023-02-15
Fix UptimeKuma url breaking (#526)

### LINKED ISSUE #526: [Bug] Uptime Kuma URL
Homer v22.10.1

When using Uptime Kuma custom service, the URL breaks when using the layout button or when typing in the search box.

The original URL changes from `uptime.server.com/status/mystatus` to `uptime.server.com/status/mystatus/status/mystatus` and each interaction with the layout button add `/status/mystatus` to the URL.

dukeofsussex:
## Description

Fix the UptimeKuma service infinitely appending the slug argument to the url.

Fixes #526 

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/UptimeKuma.vue
+++ b/src/components/services/UptimeKuma.vue
@@ -112,13 +112,15 @@ export default {
     },
   },
   created() {
-    /* eslint-disable */
-    this.item.url = `${this.item.url}/status/${this.dashboard}`;
+    // eslint-disable-next-line vue/no-mutating-props
+    this.item.url = this.item.url.endsWith(this.dashboard)
+      ? this.item.url
+      : `${this.item.url}/status/${this.dashboard}`;
     this.fetchStatus();
   },
   methods: {
     fetchStatus: function () {
-      const now = Date.now()
+      const now = Date.now();
       this.fetch(`/api/status-page/${this.dashboard}?cachebust=${now}`)
         .catch((e) => console.error(e))
         .then((resp) => (this.incident = resp));


jplanckeel-ep:
Up this subject :) We need fix :+1:

src/components/services/UptimeKuma.vue
```diff
@@ -112,13 +112,15 @@ export default {
     },
   },
   created() {
-    /* eslint-disable */
-    this.item.url = `${this.item.url}/status/${this.dashboard}`;
+    // eslint-disable-next-line vue/no-mutating-props
+    this.item.url = this.item.url.endsWith(this.dashboard)
+      ? this.item.url
+      : `${this.item.url}/status/${this.dashboard}`;
```
bastienwirtz, line 118: To avoid mutating props, I would suggest to remove all this, remove the slug property (need [documentation](https://github.com/bastienwirtz/homer/blob/main/docs/customservices.md#uptime-kuma) update too), to use the common endpoint property. If you remove those line and configure the service this way:

```yml
- name: "Uptime Kuma"
  url: "http://192.168.0.151:3001"
  endpoint: "http://192.168.0.151:3001/status/default" # Replace default by your dashboard name if any
  type: "UptimeKuma"
``` 

instead of 
```yml
- name: "Uptime Kuma"
  url: "http://192.168.0.151:3001"
  slug: "myCustomDashboard" # Defaults to "default" if not provided.
  type: "UptimeKuma"
```

It should works fine.

bastienwirtz (changes requested):
Sorry for the very very late review, thanks for spotting this and fixing it :pray:

bastienwirtz:
@jplanckeel-ep You can try to set the endpoint property in the configuration, it should work even without any code change. Something like

```yml
- name: "Uptime Kuma"
  url: "http://192.168.0.151:3001"
  endpoint: "http://192.168.0.151:3001/status/default" # Replace default by your dashboard name if any
  type: "UptimeKuma"
```

Can you confirm if it works§ Thanks :pray:

src/components/services/UptimeKuma.vue
dukeofsussex: I agree mutating props isn't great, however your suggestion doesn't seem feasible, as `endpoint` is required to be a base path for the status queries to work https://github.com/bastienwirtz/homer/blob/b168430fbf3f3a85bfcd6aa2d84aab455d4e464b/src/components/services/UptimeKuma.vue#L120-L131

eoleedi:
> @jplanckeel-ep You can try to set the endpoint property in the configuration, it should work even without any code change. Something like
> 
> ```yaml
> - name: "Uptime Kuma"
>   url: "http://192.168.0.151:3001"
>   endpoint: "http://192.168.0.151:3001/status/default" # Replace default by your dashboard name if any
>   type: "UptimeKuma"
> ```
> 
> Can you confirm if it works§ Thanks 🙏

It doesn't work.

tanasegabriel:
I think this was addressed in https://github.com/bastienwirtz/homer/pull/1057? Sorry, I missed that there was an already open PR on this, my bad!

That PR was opened based on the idea that UptimeKuma is a bit of an anti-pattern, where it didn't respect the practice of `url` + `endpoint`. While addressing that, I observed the mutate problem and fixing it was more or less a side effect of that.

Issue #632, opened 2023-04-22
Fetch status (online/offline) from uptime-kuma


tomerh2001:
As far as I am aware, the ping component is not the most versatile as it requires the end service to have a sort of health check api integrated into it.

I would like to use uptime-kuma to fetch the status of a service when I am displaying an online/offline indicator.

fbartels:
A direct integration with uptime kuma sounds nice. So far i am integrating both services through homers label feature.

bastienwirtz:
Uptime Kuma now have a dedicated integration: https://github.com/bastienwirtz/homer/blob/main/docs/customservices.md#uptime-kuma

PR #633, opened 2023-04-22
Fetch status (online/offline) from uptime-kuma

tomerh2001:
## Description

As far as I am aware, the ping component is not the most versatile as it requires the end service to have a sort of health check api integrated into it.

I would like to fetch the status of a service from uptime-kuma when I am displaying an online/offline indicator.

Fixes # (issue)
https://github.com/bastienwirtz/homer/issues/632

### DIFF
--- a/src/components/services/Ping.vue
+++ b/src/components/services/Ping.vue
@@ -9,75 +9,98 @@
 </template>
 
 <script>
-import service from "@/mixins/service.js";
-import Generic from "./Generic.vue";
+  import service from "@/mixins/service.js";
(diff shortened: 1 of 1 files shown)


bastienwirtz:
Hi @tomerh2001, 

That's a good idea but as it rely on a specific external service, it has to be in a different, separate custom service. 
Would you be open to create separate component for that?

bastienwirtz:
Uptime kuma is now available through a dedicated integration: https://github.com/bastienwirtz/homer/blob/main/src/components/services/UptimeKuma.vue.

PR #637, opened 2023-04-28
Add Transmission support

### LINKED ISSUE #634: Custom Service for Transmission
Add custom service for [Transmission](https://transmissionbt.com/) just like rTorrent or qBittorrent

jag-k:
## Description

Fixes #634

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/Transmission.vue
+++ b/src/components/services/Transmission.vue
@@ -1,0 +1,194 @@
+<template>
+  <Generic :item="item">
+    <template #content>
+      <p class="title is-4">{{ item.name }}</p>
+      <p class="subtitle is-6">
+        <span v-if="error" class="error">An error has occurred.</span>
+        <template v-else>
+          <span class="down monospace">
+            <i class="fas fa-download" aria-hidden="true"></i>
+            {{ downRate }}
+          </span>
+          <span class="up monospace">
+            <i class="fas fa-upload" aria-hidden="true"></i>
+            {{ upRate }}
+          </span>
+        </template>
+      </p>
+    </template>
+    <template #indicator>
+      <span v-if="!error" class="count"
+        >{{ count }}
+        <template v-if="count === 1">torrent</template>
+        <template v-else>torrents</template>
+      </span>
+    </template>
+  </Generic>
+</template>
+
+<script>
+import service from "@/mixins/service.js";
+import Generic from "./Generic.vue";
+
+const units = ["B", "KB", "MB", "GB"];
+
+// Take the rate in bytes and keep dividing it by 1k until the lowest
+// value for which we have a unit is determined. Return the value with
+// up to two decimals as a string and unit/s appended.
+const displayRate = (rate) => {
+  let i = 0;
+
+  while (rate > 1000 && i < units.length) {
+    rate /= 1000;
+    i++;
+  }
+  return (
+    Intl.NumberFormat(undefined, { maximumFractionDigits: 2 }).format(
+      rate || 0
+    ) + ` ${units[i]}/s`
+  );
+};
+
+export default {
+  name: "Transmission",
+  mixins: [service],
+  props: { item: Object },
+  components: { Generic },
+  data: () => ({ dl: null, ul: null, count: null, error: null }),
+  computed: {
+    downRate: function () {
+      return displayRate(this.dl);
+    },
+    upRate: function () {
+      return displayRate(this.ul);
+    },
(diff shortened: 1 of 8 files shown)


src/components/services/Transmission.vue
```diff
@@ -0,0 +1,194 @@
...
+
+      if (this.proxy?.useCredentials) {
+        options.credentials = "include";
+      }
+
+      // Each item can override the credential settings
+      if (this.item.useCredentials !== undefined) {
+        options.credentials =
+          this.item.useCredentials === true ? "include" : "omit";
+      }
```
bastienwirtz, line 130: A [custom fetch](https://github.com/bastienwirtz/homer/blob/main/src/mixins/service.js) function is available (you already included it using the `service` mixin), which handle the credential option and more. Could you update your code to use it and avoid code duplication ? Thanks :pray:

src/components/services/Transmission.vue
```diff
@@ -0,0 +1,194 @@
...
+      } else {
+        url = this.endpoint;
+        if (url.endsWith("/")) {
+          url = url.slice(0, -1);
+        }
+        if (url.endsWith("/web")) {
+          url = url.slice(0, -4);
+          url += "/rpc";
+        }
+      }
```
bastienwirtz, line 156: We don't need another config key for `/rpc` endpoint. the `endpoint` config key is made for this. It's supposed to be the url where to fetch the data. You just have to write in the documentation that `endpoint` should contain the rpc url.

src/components/services/Transmission.vue
```diff
@@ -0,0 +1,194 @@
...
+
+      const response = await fetch(url, options);
```
bastienwirtz, line 158: use `this.fetch` here.

src/components/services/Transmission.vue
```diff
@@ -0,0 +1,194 @@
...
+
+      // If the session id is not set, we need to get it from the response
+      if (response.status === 409 && !this.sessionId) {
+        const body = await response.text();
+        const match = body.match(/X-Transmission-Session-Id: ([a-z0-9]+)/i);
```
bastienwirtz, line 163: I think the X-Transmission-Session-Id exist in the response headers, it would be easier / cleaner to get it from the headers.

bastienwirtz (changes requested):
Thanks for your contribution @jag-k. A transmission custom service will be nice! 
I think we can simplify a bit the data fetching part, but other than than, it should be good.

src/components/services/Transmission.vue
thedroid: @bastienwirtz 

With the existing custom fetch, we don't have access to the response in the event of a non Response.OK result.  We would need access to the response in order to obtain the  "X-Transmission-Session-Id" after we are sent a 409 status code.

I have hacked it trivially with, but I'm sure there is a better / more correct way.

In service.js
```
`fetch: function (path, init, json = true, forceResponse = false)`
...
...
        if (!response.ok) {
          if (!forceResponse)
            throw new Error("Not 2xx response");
          else
            return response;
        }
```

src/components/services/Transmission.vue
royto: You can use `response.headers.get('x-transmission-session-id')` to get the session-id

SenorSmartyPants:
I'm testing this PR, but I'm stuck on enabling CORS. I have transmission RPC enabled (works with android remote apps), but I don't see a way to enable CORS natively in transmission. Are you putting it behind a proxy to enable this?

SenorSmartyPants:
I got CORS working on NGINX for this by following [this](https://enable-cors.org/server_nginx.html) and then modifying the `Access-Control-Allow-Headers` in the OPTIONS section to add `X-Transmission-Session-Id`

`add_header 'Access-Control-Allow-Headers' 'X-Transmission-Session-Id,DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';`

I need to do that to get my version 3.x transmission to work. But on my other transmission install that was already 4.x I didn't need CORS at all and I could communicate without a proxy. I ended up upgrading my 3.x install to 4.x

Thank you for this PR.

SenorSmartyPants:
I added some changes to this as a PR into jag-k repo

https://github.com/jag-k/homer/pull/2

bastienwirtz:
Transmission is now natively available in the latest version.

PR #650, opened 2023-05-21
Add Nextcloud custom service

meonkeys:
## Description

This adds support for Nextcloud custom services. My motivation is so I can use this with my own Nextcloud server. This adds no new dependencies.

## Type of change

## Checklist:


src/components/services/Nextcloud.vue
```diff
@@ -0,0 +1,106 @@
...
+    status: function () {
+      if (this.fetchOk) {
+        if (this.maintenance) {
+          return "maintenance";
+        } else {
+          return "online";
+        }
+      } else {
+        return "offline";
+      }
```
bastienwirtz, line 48: I think this part could be simplified like this: 
```suggestion
      if (!this.fetchOk) {
        return "offline";
      }
      return this.maintenance ? "maintenance" : "online";
```

bastienwirtz (changes requested):
Hello @meonkeys, 
Thanks for this contribution! 
A little thing but otherwise it look good :+1:

bastienwirtz (approved):
Sorry for the very very late re-review, everything looks good !

PR #651, opened 2023-05-22
Allow setting no-cors in ping service

kmjayadeep:
## Description

Currently Ping custom service does not work if the service is hosted under a different domain and no `Access-Control-Allow-Origin`  header is set on the api response. This PR fixes this issue by allowing to make requests with `mode: no-cors` in fetch api.

More info about the `mode` field can be found [here](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode)

## Type of change

## Checklist:

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -144,6 +144,8 @@ For Ping you need to set the type to Ping and provide a url. By default the HEAD
   method: "head"
 ```
 
+You can also specify an optional field `mode` as `no-cors` to allow cross-origin requests from the browser. Setting this field will show the status of the service as `online` if the request is successful, irrespective of the response status code. Check the official [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Request/mode) for more information about this field.
+
 ## Prometheus
 
 For Prometheus you need to set the type to Prometheus and provide a url.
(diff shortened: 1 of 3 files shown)


DanielBaulig:
@bastienwirtz Any chance we can get this merged? This would give us a great alternative to setting up proxy servers for fixing CORS related ping issues.

bastienwirtz:
Hello @kmjayadeep !

Thanks for looking into this. I agree CORS make this custom service use case very limited, however, sadly, this solution does not work. 
The `no-cors` mode prevent JS from accessing any header, including response status. It means that is the target site returns a http 500 error code (or any error code, even just a 404 not found), it will be displayed as online, when it's not. **The card will always show an online status**. 

You can test it with http://httpstat.us/500 , http://httpstat.us/404, ... it's displayed as online.
I'm closing the PR as I don't see any other possibility, but feel free to reopen if you have other ideas.

kmjayadeep:
Hi @bastienwirtz 
Thank you for your reply!

> The no-cors mode prevent JS from accessing any header, including response status. It means that is the target site returns a http 500 error code (or any error code, even just a 404 not found), it will be displayed as online, when it's not. The card will always show an online status.

This is indeed the expected behavior, which I also explained in the readme. I think it is still an improvement considering that the alternative is to not show a status at all. Sadly there are many services which we don't have control over to allow cors from homer,  and is the only way around it.

PR #694, opened 2023-08-30
Use SUBFOLDER as asset base to allow mounting homer at a different base path.

### LINKED ISSUE #578: Hosting homer behind reverse proxy on subpath
I'm trying to use Traefik to host Homer behind a Reverse Proxy. The path I'm trying to use is 192.168.2.207/homer. However all .js includes are being done like this:
image
```
<link rel="stylesheet" href="./resources/index41a7d45b.css">
```

A way to host homer on a path lower then the root of a webshite.

I tried all kinds of things with Traefik but haven't been able to resolve it. The only other option is trying to do something like homer.blah rather then blah/homer.

**Additional context**
N/A

ignatz:
To try run e.g.:

    $ docker build -t homer --build-arg SUBFOLDER=/bar . && docker run --rm -p 8080:8080 -e SUBFOLDER=/bar homer

before and after. Previously, the base path wasn't set. So when browsing to the lighthttp alias, you'd be instantly be sent back to the root path.

## Description

Fixes #578 

## Type of change

## Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,4 @@ dockerfile
 .git
 screenshot.png
 node_modules
+dummy-data
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 # build stage
 FROM node:lts-alpine as build-stage
 
+ARG SUBFOLDER ""
+
 WORKDIR /app
 
 COPY package*.json ./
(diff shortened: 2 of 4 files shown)


ignatz:
Note that a downside of this relatively straight forward approach is that users will have to build their own docker package with the build_arg set.

bastienwirtz:
Hello, 
Thanks for looking into this 🙂
Having users to build homer is not ideal, and using the `SUBFOLDER` variable as described in the documentation just works, I don't think we need more. Hope that works for you.

PR #709, opened 2023-10-06
feat: Authentication on qBittorrent

garretaserra:
Allow optional authentication on qBittorrent

## Description

Currently the only way to add qBittorrent is to configure it so that Homer's IP bypasses authentication.
Instead we can easily authenticate with the API with a request and cookie based authentication. The only problem is that because of CSRF qBittorrent will only allow API requests from its own website. If CSRF is disabled you can authenticate correctly. In [this](https://github.com/qbittorrent/qBittorrent/issues/17598) conversation (which I have included in the docs so that if users whant to know what this is) its talked about how this is not oficially supported.

Fixes # (issue)

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] I've read & comply with the [contributing guidelines](https://github.com/bastienwirtz/homer/blob/main/CONTRIBUTING.md)
- [x ] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. 
- [x ] I have made corresponding changes to the documentation (README.md).
- [ x] I've checked my modifications for any breaking changes, especially in the `config.yml` file

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -324,8 +324,7 @@ This service displays the global upload and download rates, as well as the numbe
 listed. The service communicates with the qBittorrent API interface which needs
 to be accessible from the browser. Please consult
 [the instructions](https://github.com/qbittorrent/qBittorrent/pull/12579)
-for setting up qBittorrent and make sure the correct CORS-settings are applied. Examples for various
-servers can be found at [enable-cors.org](https://enable-cors.org/server.html).
+for setting up qBittorrent and make sure the correct CORS-settings are applied, for example by setting the header: "`Access-Control-Allow-Origin: *`. Examples for various servers can be found at [enable-cors.org](https://enable-cors.org/server.html). qBittorrent [doesn't support](https://github.com/qbittorrent/qBittorrent/issues/17598#issuecomment-1225538387) authenticating from other websites. Nevertheless if you know what you are doing you can disable CSRF protection and it will work.
 
 ```yaml
 - name: "qBittorrent"
@@ -335,6 +334,10 @@ servers can be found at [enable-cors.org](https://enable-cors.org/server.html).
   rateInterval: 2000 # Interval for updating the download and upload rates.
   torrentInterval: 5000 # Interval for updating the torrent count.
   target: "_blank" # optional html a tag target attribute
+
+  # Optional authentication, requires disabling CSRF protection
+  username: "username"
+  password: "password"
 ```
 
 ## Copy to Clipboard
--- a/src/components/services/qBittorrent.vue
+++ b/src/components/services/qBittorrent.vue
@@ -72,6 +72,11 @@ export default {
       setInterval(() => this.fetchCount(), torrentInterval);
     }
 
+    // authenticate if username and password are provided
+    if (this.item.username != undefined && this.item.password != undefined) {
+      this.authenticate(this.item.username, this.item.password);
+    }
+
     this.getRate();
     this.fetchCount();
   },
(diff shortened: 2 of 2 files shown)


rifatgungor (approved):

caarlos0 (approved):

caarlos0:
This seems like it'll do the authentication request, but will not use the resulting cookie in the following requests... so the issue remains:

CleanShot 2024-01-04 at 00 46 24@2x
CleanShot 2024-01-04 at 00 46 50@2x

caarlos0:
```diff
diff --git a/docs/customservices.md b/docs/customservices.md
index 2dfdceb..18f60a8 100644
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -324,8 +324,7 @@ This service displays the global upload and download rates, as well as the numbe
 listed. The service communicates with the qBittorrent API interface which needs
 to be accessible from the browser. Please consult
 [the instructions](https://github.com/qbittorrent/qBittorrent/pull/12579)
-for setting up qBittorrent and make sure the correct CORS-settings are applied. Examples for various
-servers can be found at [enable-cors.org](https://enable-cors.org/server.html).
+for setting up qBittorrent and make sure the correct CORS-settings are applied, for example by setting the header: "`Access-Control-Allow-Origin: *`. Examples for various servers can be found at [enable-cors.org](https://enable-cors.org/server.html). qBittorrent [doesn't support](https://github.com/qbittorrent/qBittorrent/issues/17598#issuecomment-1225538387) authenticating from other websites. Nevertheless if you know what you are doing you can disable CSRF protection and it will work.
 
 ```yaml
 - name: "qBittorrent"
@@ -335,6 +334,10 @@ servers can be found at [enable-cors.org](https://enable-cors.org/server.html).
   rateInterval: 2000 # Interval for updating the download and upload rates.
   torrentInterval: 5000 # Interval for updating the torrent count.
   target: "_blank" # optional html a tag target attribute
+
+  # Optional authentication, requires disabling CSRF protection
+  username: "username"
+  password: "password"
 ```
 
 ## Copy to Clipboard
diff --git a/src/components/services/qBittorrent.vue b/src/components/services/qBittorrent.vue
index d47ed0b..4259b79 100644
--- a/src/components/services/qBittorrent.vue
+++ b/src/components/services/qBittorrent.vue
@@ -72,13 +72,20 @@ export default {
       setInterval(() => this.fetchCount(), torrentInterval);
     }
 
+    // authenticate if username and password are provided
+    if (this.item.username != undefined && this.item.password != undefined) {
+      this.authenticate(this.item.username, this.item.password);
+    }
+
     this.getRate();
     this.fetchCount();
   },
   methods: {
     fetchCount: async function () {
       try {
-        const body = await this.fetch("/api/v2/torrents/info");
+        const body = await this.fetch("/api/v2/torrents/info", {
+          credentials: 'include'
+        });
         this.error = false;
         this.count = body.length;
       } catch (e) {
@@ -88,7 +95,9 @@ export default {
     },
     getRate: async function () {
       try {
-        const body = await this.fetch("/api/v2/transfer/info");
+        const body = await this.fetch("/api/v2/transfer/info", {
+          credentials: 'include'
+        });
         this.error = false;
         this.dl = body.dl_info_speed;
         this.ul = body.up_info_speed;
@@ -97,6 +106,22 @@ export default {
         console.error(e);
       }
     },
+    authenticate: async function (username, password) {
+      try {
+        const reqParams = {
+          username: username,
+          password: password,
+        };
+        const body = await this.fetch("/api/v2/auth/login", {
+          method: "POST",
+          body: new URLSearchParams(reqParams),
+        });
+        this.error = false;
+      } catch (e) {
+        this.error = true;
+        console.error(e);
+      }
+    },
   },
 };
 </script>
```

this fixes it for me (because both homer and qbittorrent are in the same domain, but different ports)... ideally I think we should use a cookiejar or something...

garretaserra:
@caarlos0 thanks for the feedback, for me it actually works without the modification you proposed.
Nevertheless, I agree that it is not a solution but rather a bad workaround. I'm not familiar with "cookiejar" but from what I could understand, it would only work as a separate service from the frontend, which is Homer.

After some research I think this is not feasible due to how qbittorrent is built. Another workaround that I have used on previous projects with a similar problem to this is just to have a proxy handle the authentication for the website but again its not the best solution and most likely not worth it for Homer.

In conclusion I think that either qBittorrent needs to change to support this or Homer would need to have a backend to do the auth from there.

PR #735, opened 2024-01-12
feat: Import secretes from secrets.yml

### LINKED ISSUE #609: Passing secrets or env vars to the config.yaml

Hey! Thanks for this project!

Is there a way to pass secrets into `config.yaml` file via some sort of keyword?

For example, Home Assistant and zigbee2mqtt projects provide an ability to pass secrets or env vars via keywords in a YAML files. Like this:

```yaml
token: !env TOKEN_ENV_VARIABLE
# or
token: !secret token # token field from secrets.yaml file
```

¯\_(ツ)_/¯ 

**Additional context**

I am open sourcing my [smart home config](https://github.com/denysdovhan/smart-home) and would like to use Homer for a dashboard of my smart home server. I would love to publish my config, by I also need to pass some secrets to the config (like token for Portainer or Pi Hole, etc)

flooxo:
## Description

I have implemented a new function so that it is now possible to import secrets from a `secrets.yml` file using the keyword `!secret`.

When loading the config, the corresponding key is loaded with the token from the secrets.yml file and replaced everywhere in the `config.yml` file

Fixes #609

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -194,3 +194,29 @@ The `/assets/manifest.json` can also be edited to change the app (pwa) name, des
 ## PWA Icons
 
 See icons documentation [here](https://github.com/bastienwirtz/homer/blob/main/public/assets/icons/README.md).
+
+
+## Importing Secrets
+
+You can import sensitive information such as API keys or passwords into your `config.yml` file securely using the `!secret` keyword. The corresponding tokens are stored in the `secrets.yml` file. This approach allows you to keep confidential information separate, eliminating the need to expose sensitive information directly in your configuration.
+
+
+### Example:
+
+In your `secrets.yml` file:
+
+```yaml
+api_key: your_super_secret_api_key
+db_password: your_secure_database_password
+```
+
+In your `config.yml` file:
+
+```yaml
+services:
+  - name: "Example Service"
+    api_key: !secret api_key
+    database:
+      password: !secret db_password
+# ...
+```
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,6 @@ yarn-error.log*
 
 # App configuration
 config.yml
+secrets.yml
 
 .drone.yml
--- a/public/assets/config-demo.yml.dist
+++ b/public/assets/config-demo.yml.dist
@@ -83,7 +83,7 @@ services:
         node: "node1"
         warning_value: 50
         danger_value: 80
-        api_token: "xxxxxxxxxxxx"
+        api_token: !secret proxmox_api
       - name: "An awesome app"
         logo: "assets/tools/sample.png"
         subtitle: "Bookmark example"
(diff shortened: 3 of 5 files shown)


kgromer:
This is awesome! Please pull this into the main branch.

docs/configuration.md
```diff
@@ -194,3 +194,29 @@ The `/assets/manifest.json` can also be edited to change the app (pwa) name, des
 ## PWA Icons
 
 See icons documentation [here](https://github.com/bastienwirtz/homer/blob/main/public/assets/icons/README.md).
+
+
+## Importing Secrets
+
+You can import sensitive information such as API keys or passwords into your `config.yml` file securely using the `!secret` keyword. The corresponding tokens are stored in the `secrets.yml` file. This approach allows you to keep confidential information separate, eliminating the need to expose sensitive information directly in your configuration.
```
bastienwirtz, line 201: This secret file is as public as the main config file. Having it separated can help manage the secrets, but it does not add any security.

bastienwirtz (commented):
Hello, 

Indeed, secret in homer config file can be problematic. If you use secrets, Homer should not be exposed publicly (or behind some kind of auth proxy). 
However, this approach can be misleading, users could think their secrets are secure in the separate file and expose it to the internet without any auth. It also add an additional request for no real value. 

The only way to hide credentials would be to proxy the request on the server side.

flooxo:
Thank you for the feedback. I certainly agree, it does not make the secrets more secure. Perhaps the wording in the documentation should be revised.

The advantage of this feature is that you can publish your config without worrying, as discussed in #609.
And I would say that most homer users also run other services and therefore know how to deal with secrets.  This feature gives them a more structured way to manage secrets without enforcing specific security implementations :)

crabdancing:
Just ran into this for NixOS. Just wanted to +1 the importance of this on declarative distros. There are many security models where it still makes sense to do this.

NedWilbur:
An option of getting secrets from env variables would be more ideal IMO

bastienwirtz:
@NedWilbur As a client side app, it's not possible for Homer to read env vars, **but** you can use a server side buddy like [`Corsair`](https://github.com/bastienwirtz/corsair), to inject the credentials from env vars.

instantdreams:
I keep all my server configurations in repositories. My compose files always reference a `.env` file and also a `.env.secrets` file if needed. The directory structure for all containers that use secrets looks like this:
   * compose.yaml
   * .env.example - standard environment variables, expected not to change
   * .env.secrets.example - api tokens, passwords, secrets

In my repositories the secrets are all placeholders, such as `PIHOLEAPIKEY=[piholeapikey]`

When I prepare the instance to be stood up, I do the following:

```
cp .env.example .env
cp .env.secrets.example .env.secrets
nano .env.secrets
docker compose pull
docker compose up --detach
```

I edit the `.env.secrets` file to replace the placeholders with the actual tokens and passwords so they are only on the server.

I would love for Homer to support a substitution process like in this commit which would substitute the secrets in the file.

Issue #726, opened 2023-12-05
Different URLs based on where you open homer


Cyberbeni:
**Is your feature request related to a problem? Please describe.**
-
I'm accessing my homer dashboard from both my computer and my iOS/iPadOS devices from both my home network and through tailscale. It would be nice to have different urls based on those conditions.

**Describe the solution you'd like**
-
Links with same host and different port, so if I want to open HomeAssistant on port 8123, it uses either 192.168.x.y or myserver.tailscale.net as the host (whichever I'm using to access the page) and gets the rest of the url from the config. ([relevant SO question I've found](https://stackoverflow.com/questions/6016120/relative-url-to-a-different-port-number-in-a-hyperlink))

It would be also nice if iOS could have the deeplink to the iOS app for HomeAssistant ( https://home-assistant.io/ios/ ) and everywhere else would have the url for my local server. (And probably another configuration for Android)

**Describe alternatives you've considered**-
-

**Additional context**
-

glebkrapivin:
+1 

I have exactly the same request

glebkrapivin:
@Cyberbeni 

Hey!

I am by no means  a js coder, but didn't wanna wait. 

If you navigate to `/src/components/Service.vue` and add the following two lines, then `docker build...` - it would work :) 

```
<template>
  <component :is="component" :item="item" :proxy="proxy"></component>
</template>

<script>
import {defineAsyncComponent} from "vue";
import Generic from "./services/Generic.vue";

export default {
  name: "Service",
  props: {
    item: Object,
    proxy: Object,
  },
  computed: {
    component() {

      const type = this.item.type || "Generic";
      if (type === "Generic") {
       # start here
        if (this.item.url.startsWith('/') || this.item.url.startsWith(':')) {
          this.item.url = window.location.protocol + '//' + window.location.hostname + this.item.url
       }
       # end here 
        return Generic;
      }
      return defineAsyncComponent(() => import(`./services/${type}.vue`));
    },
  },
};
</script>
```

PR #741, opened 2024-01-20
Allow using the same host with different port for urls

Cyberbeni:
## Description

Some places still not have this modification, for example the links section, as I don't know how to implement this url calculation in multiple places nicely in Vue/JS.
I've only had a problem with trailing slash on iOS, not on desktop.
I've also had a problem with having to recreate the "Add to Home" icon for changes to take effect on iOS. Maybe trailing slash wasn't even a problem and not even the first iteration of the code had any effect. (But then why did the config change had an effect?) -- Might be unrelated.

Partial fix for #726 

## Type of change

## Checklist:

### DIFF
--- a/src/components/Service.vue
+++ b/src/components/Service.vue
@@ -14,6 +14,13 @@ export default {
   },
   computed: {
     component() {
+      if (this.item.url.startsWith('/') || this.item.url.startsWith(':')) {
+        let hostname = window.location.hostname;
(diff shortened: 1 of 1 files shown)


glebkrapivin:
Thanks for making this a PR :) 

And you are right about the trailing slash, I've noticed a similar problem on my iphone once

PR #778, opened 2024-05-16
Emby/Jellyfin: Add moviesandseries option

### LINKED ISSUE #681: Jellyfin: show both movies and series counter?
Right now one can choose to either show `libraryType: "movies"` or `"series"` on a Jellyfin tile. Whereas the "series" option is showing the total number of episodes and the total number of series on your Jellyfin.

I would like to see a third option, which is displaying the total number of movies and the total number of series. In the end this should occupy the same amount of space as `"series"` (episodes/series) is right now.

For one the `"series"` option could be split into `"episodes"` and` "series"` and then it could be made possible to chain those in one line of `libraryType:`. So for example `libraryType: "episodes", "movies"` which would display the total number of episodes and the number of movies.

Screenshot for reference.
image

SenorSmartyPants:
Will display counts of movies and series

Fixes #681

### DIFF
--- a/src/components/services/Emby.vue
+++ b/src/components/services/Emby.vue
@@ -48,6 +48,8 @@ export default {
         return `${this.movieCount} movies`;
       else if (this.item.libraryType === "series")
         return `${this.episodeCount} eps, ${this.seriesCount} series`;
+      else if (this.item.libraryType === "moviesandseries")
+        return `${this.movieCount} movies, ${this.seriesCount} series`;
       else return `wrong library type 💀`;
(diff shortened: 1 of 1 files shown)


emilywotruba (approved):
It's a minor yet wanted change, can we get this merged?

cmatute7712:
Hey folks, I put up a pull request that does something similar to this one but allows for multiselect across any of the three library types. FYI. [#785](https://github.com/bastienwirtz/homer/pull/785)

PR #783, opened 2024-05-20
Adding support for Home Assistant

emilywotruba:
## Description
Created new custom service for Home Assistant and added documentation.
Allows displaying the name and version of the instance, as well as the count of entities:

screenshot

Fixes # N/A

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/HomeAssistant.vue
+++ b/src/components/services/HomeAssistant.vue
@@ -93,7 +93,7 @@ export default {
     fetchServerStatus: async function () {
       const headers = this.headers;
 
-      this.fetch("/api/", { headers })
+      return this.fetch("/api/", { headers })
         .then((response) => {
           if (response && response.message) this.status = "running";
           else throw new Error();
--- a/src/components/services/HomeAssistant.vue
+++ b/src/components/services/HomeAssistant.vue
@@ -1,0 +1,165 @@
+<template>
+  <Generic :item="item">
+    <template #content>
+      <p class="title is-4">{{ item.name }}</p>
+      <p class="subtitle is-6">
+        <template v-if="item.subtitle">
+          {{ item.subtitle }}
+        </template>
+        <template v-else>
+          {{ details }}
+        </template>
+      </p>
+    </template>
+    <template #indicator>
+      <div v-if="status" class="status" :class="status">
+        {{ status }}
+      </div>
+    </template>
+  </Generic>
+</template>
+
+<script>
+import service from "@/mixins/service.js";
+import Generic from "./Generic.vue";
+
+export default {
+  name: "HomeAssistant",
+  mixins: [service],
+  props: {
+    item: Object,
+  },
+  components: {
+    Generic,
+  },
+  data: () => ({
+    status: "",
+    msg: "",
+    version: "",
+    entities: 0,
+  }),
+  computed: {
+    headers: function () {
+      return {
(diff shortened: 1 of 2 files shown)


src/components/services/HomeAssistant.vue
```diff
@@ -0,0 +1,165 @@
...
+      if (!this.item.separator) this.item.separator = " ";
+      if (!this.item.logo) this.item.logo = `${this.item.url}/static/icons/favicon-192x192.png`;
+      this.fetchServerStats();
+    }
+  },
+  methods: {
+    fetchServerStatus: async function () {
+      const headers = this.headers;
+
+      this.fetch("/api/", { headers })
```
bastienwirtz, line 96: You could return the fetch call to make possible to await this.
```suggestion
      return this.fetch("/api/", { headers })
```

src/components/services/HomeAssistant.vue
```diff
@@ -0,0 +1,165 @@
...
+        }
+      }
+      
+      return details.join(separator);
+    },
+  },
+  created() {
+    this.fetchServerStatus();
+
+    if (!this.item.subtitle && this.status !== "dead") {
```
bastienwirtz, line 85: fetchServerStatus is async, and this.status will not have the server status yet. You either have to await or use the promise (.then(....)) to make it reliable.

bastienwirtz (changes requested):
Thanks for your contribution @emilywotruba!
A small issue to fix on the data fetching and we should be good to go.

src/components/services/HomeAssistant.vue
emilywotruba: You're right, i used `then` to resolve this.

emilywotruba:
Thank you for reviewing my request @bastienwirtz. Are your concerns addressed by my new commits?

src/components/services/HomeAssistant.vue
```diff
@@ -0,0 +1,165 @@
...
+      
+      return details.join(separator);
+    },
+  },
+  created() {
+    this.fetchServerStatus().then(() => {
+      if (!this.item.subtitle && this.status !== "dead") {
+        if (!this.item.items) this.item.items = ["name", "version"];
+        if (!this.item.separator) this.item.separator = " ";
+        if (!this.item.logo) this.item.logo = `${this.item.url}/static/icons/favicon-192x192.png`;
```
bastienwirtz, line 87: ESlint is not happy with the prop mutation (cf: https://github.com/bastienwirtz/homer/actions/runs/9179879689/job/25243741529?pr=783#step:6:12), I forgot about that in my first review, sorry. Could you see to use local variable or data variables for example? thanks!

bastienwirtz (commented):
Aside from the linting issue, everything looks good @emilywotruba, thanks :pray:

src/components/services/HomeAssistant.vue
emilywotruba: Ah sad, I really liked the idea of automatically grabbing the icon of home assistant if no logo was provided. I will fix it, thank you for your patience.

bastienwirtz: Yeah that was a good idea! (could be applied to other service too. Thinking about it, in this case it should be harmless, we could disable the lint rules :thinking: 
I'll look into it, we can merge it like that for now, and change it back in the future if we are ok with the prop mutation.

bastienwirtz (approved):

PR #784, opened 2024-05-22
Adding customization options to PiHole custom service

emilywotruba:
## Description

Currently, the PiHole custom service can only show the percentage of blocked queries. I was interested in other statistics, so I added two new items to its configuration:
- `items` an array of keys specifying in which fields of the api you are interested in
- `format` a string giving the format of the subtitle. Here, `"{n}"` will be replaced by the value of the `n`'th key from `items`

**Examples:** If we leave `items` and `format` empty, the old subtitle will be generated, so this addition does not break old configurations. This is equivalent to setting:
```yaml
items: ["ads_percentage_today"]
format: "{0}% blocked"
```
grafik

Another interesting statistic:

```yaml
items: ["queries_forwarded", "dns_queries_today"]
format: "{0} of {1} queries forwarded"
```
grafik

The second addition is that instead of a red `disabled`, the status will now be an orange `dns only` if blocking is off. This is because PiHole itself can be enabled while the blocking feature is disabled:
grafik

Fixes # N/A

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -6,8 +6,8 @@
         <template v-if="item.subtitle">
           {{ item.subtitle }}
         </template>
-        <template v-else-if="percentage">
-          {{ percentage }}&percnt; blocked
+        <template v-else>
+          {{ details }}
         </template>
       </p>
     </template>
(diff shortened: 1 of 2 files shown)


bastienwirtz (approved):

src/components/services/PiHole.vue
```diff
@@ -34,30 +34,53 @@ export default {
...
+            } else {
+              this.status_msg = "dns only";
+            }
 
-      this.status = result.status;
-      this.ads_percentage_today = result.ads_percentage_today;
+            for (const i in this.items)
+              this.results[i] = response[this.items[i]];
+          } else throw new Error();
+          1;
```
bastienwirtz, line 78: Is this on purpose ?

bastienwirtz (commented):
Thanks for this @emilywotruba ! Just one question and we should be good to merge. 
Also if you could rebase your branch from main, it would be perfect :pray:

PR #785, opened 2024-06-01
Imeplements multi selector for library type for Emby/Jellyfin. Allows…

### LINKED ISSUE #681: Jellyfin: show both movies and series counter?

cmatute7712:
homer-jellyfin-multi-library

## Description

Imeplements multi selector for library type for Emby/Jellyfin. Allows for comma/space separated values as input rather than matching on equals. Also updated docs to reflect change

Fixes #681 

## Type of change

## Checklist:

 ͡° ͜ʖ ͡°

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -203,7 +203,7 @@ You need to set the type to Emby, provide an api key and choose which stats to s
   url: "http://192.168.0.151/"
   type: "Emby"
   apikey: "<---insert-api-key-here--->"
-  libraryType: "music" #Choose which stats to show. Can be one of: music, series or movies.
+  libraryType: "music, movies, series" #Choose which stats to show. Can be one of: music, series or movies or any combination of the three separted by commas or spaces
 ```
 
 ## Uptime Kuma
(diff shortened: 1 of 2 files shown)


emilywotruba:
You use a string of comma separated keywords instead of an array of strings, why? Compare to the HomeAssistant's component multi selector, which uses an array that also determines order of elements. Other than that I like it.

cmatute7712:
@emilywotruba I had't thought of that, wonderful idea. Will monkey around and see what I come up with. Keep an eye out :)

cmatute7712:
Will re-open with array of strings over string splitting

Issue #721, opened 2023-11-09
Mealie Custom Service is not working


DevFreezing:
**Describe the bug**
Mealie Custom Service is not working, Homer does not connect to the API. 

**Expected behavior**
Should show the number of recipes Mealie is keeping organized or the planned meal for today if one is planned

**Logs & errors**
I suspect that is s due to the URLs of the API are not properly set in the file src/components/services/Mealie.vue
For example at line 47 is calling /api/meal-plans/today/ but if we look at the mealie [API documentation ](https://nightly.mealie.io/api/redoc/)the URL is api/groups/mealplans/today

**Screenshots**

**Configuration**
My yml for mealie is:
 `  - name: "Applications"
    icon: "fas fa-cloud"
    items:
      - name: "Mealie"
        logo: "assets/icons/mealie.png"
        type: "Mealie"
        keywords: "self hosted recipe manager "
        url: "http://10.10.1.102:9000"
        target: "_blank" # optional html a tag target attribute
        apikey: "XXXX"`

```

bobonthenet:
I'm experiencing issues with this too. I tried changing the urls to match the api docs, but it still isn't working.

bobonthenet:
Turns out my Nginx config had issues too which is why changing the urls initially didn't work. I've created a PR to update those urls.

JohWQ:
I'm also experiencing issues with this.

The Homer docker logs don't even mention anything to do with Mealie.

Inside of the docker container i can reach the domain with nslookup: mealie.domain.xyz.

PR #819, opened 2024-10-31
Fixes mealie when meal plan is empty

rhardih:
If there's nothing in the meal plan for the current day, the /api/groups/mealplans/today endpoint will simply return an `[]` response.

This case wasn't guarded for in the template before this fix.

## Description

Fixes # (issue)

Follow up to https://github.com/bastienwirtz/homer/issues/721.

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/Mealie.vue
+++ b/src/components/services/Mealie.vue
@@ -6,7 +6,9 @@
         <template v-if="item.subtitle">
           {{ item.subtitle }}
         </template>
-        <template v-else-if="meal"> Today: {{ meal[0].recipe.name }} </template>
+        <template v-else-if="meal">
+          Today: {{ (meal && meal.length > 0) ? meal[0].recipe.name : 'Nothing planned' }}
+        </template>
         <template v-else-if="stats">
           happily keeping {{ stats.totalRecipes }} recipes organized
         </template>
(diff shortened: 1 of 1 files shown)


src/components/services/Mealie.vue
```diff
@@ -6,7 +6,9 @@
         <template v-if="item.subtitle">
           {{ item.subtitle }}
         </template>
-        <template v-else-if="meal"> Today: {{ meal[0].recipe.name }} </template>
+        <template v-else-if="meal">
+          Today: {{ (meal && meal.length > 0) ? meal[0].recipe.name : 'Nothing planned' }}
```
bastienwirtz, line 10:
```suggestion
          Today: {{ meal.length > 0 ? meal[0].recipe.name : 'Nothing planned' }}
```
Because it's already checked on line 9 ?  
Maybe better: Use a computed property to put the logic away from the template?

bastienwirtz (commented):
Good catch @rhardih! Thanks :pray:

rhardih:
@bastienwirtz I've moved the texts to be computed properties as well now. 👍

bastienwirtz (approved):

bastienwirtz:
Thank you @rhardih !

PR #825, opened 2024-11-13
Ping service timeout

### LINKED ISSUE #804: Ping-Service should use a timeout
By default, the Ping-Service uses `fetch` to get the status of the target web application

```
      this.fetch("/", { method, cache: "no-cache" }, false)
        .then(() => {
          this.status = "online";
        })
        .catch(() => {
          this.status = "offline";
        });
```

If a web application is offline, and there is no response, fetch waits a few seconds until it times out.
Until there was no timeout, the web application will not be marked as offline.

Is it possible to add a "timeout" option (in addition to the already existing method option) to set the timeout to a given value.
By setting it for example to 500ms, the web application would be marked as "offline" much faster.

Canti23:
## Description

Added support for the timeout option for the Ping service. The default value is 2000 ms, which should be sufficient for most services.

Fixes #804

## Type of change

## Checklist:

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -153,7 +153,7 @@ This card checks if the target link is available. All you need is to set the `ty
   logo: "assets/tools/sample.png"
   url: "https://www.wikipedia.org/"
   # method: "head"
-  # timeout: 500
+  # timeout: 500 # in ms. default 2000
   # subtitle: "Bookmark example" # By default, request round trip time is displayed when subtitle is not set.
 ```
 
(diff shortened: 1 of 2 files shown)


docs/customservices.md
```diff
@@ -153,6 +153,7 @@ This card checks if the target link is available. All you need is to set the `ty
   logo: "assets/tools/sample.png"
   url: "https://www.wikipedia.org/"
   # method: "head"
+  # timeout: 500
```
bastienwirtz, line 156:
```suggestion
  # timeout: 500 # in ms. default 2000
```

bastienwirtz (changes requested):
Small suggestion regarding the doc, aside from that, it's all good. Thanks @Canti23 !

bastienwirtz (approved):

Canti23:
Thanks, thought about that, but haven't figured out how to write it succinctly :)

Issue #130, opened 2020-09-12
Using SSL with Homer


jtovar43:
I was just wondering how I would go about adding SSL to the Dashboard. Thanks!

fbartels:
Are you using the Docker image? If yes then you would/could add a reverse proxy to provide ssl. If you are just using the static assets then Homer is just like any other static website (choose your own webserver and secure it with ssl).

jtovar43:
Thanks! I got it working with a reverse proxy. Didn't know you could do that.

J7mbo:
Thanks @Alevsk . This still works in 2026. I don't want to "just use a reverse proxy" because I want true End-to-End L7 Encryption in my cluster and don't want to terminate at say Traefik. Appreciate you finding this!

PR #848, opened 2024-12-27
allow using SSL

VergilGao:
## Description

Add an option to use SSL.
The inspiration for this PR comes from this issue https://github.com/bastienwirtz/homer/issues/130

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,8 @@ ENV GID=1000 \
     PORT=8080 \
     SUBFOLDER="/_" \
     INIT_ASSETS=1 \
-    IPV6_DISABLE=0
+    IPV6_DISABLE=0 \
+    SSL_ENABLE=0
 
 RUN addgroup -S lighttpd -g ${GID} && adduser -D -S -u ${UID} lighttpd lighttpd && \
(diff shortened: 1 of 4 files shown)


VergilGao:
My Nginx server is running on another computer. This PR can effectively use the SSL encryption Nginx server and the HOMER instance.

thibaut1304:
👋 Hello! I just have a quick question.

Instead of embedding SSL support directly into Homer, wouldn’t it be simpler and more modular to recommend putting Homer behind a reverse proxy like Traefik or Nginx, which would handle SSL/TLS certificates automatically (e.g. Let’s Encrypt)?

That way:
  - Homer stays lightweight and frontend-only,
  - SSL is centralized and easier to maintain,
  - No need to configure or bundle certificates directly in the app.

Personally, I use this approach for all my services and find it really simplifies setup and maintenance. Just wondering what you think! 🙂

bastienwirtz:
Hello !

As @thibaut1304 pointed out, the easiest way to achieve this is using a reverse proxy. You will get much more flexibility on the configuration. 

I'm closing the issue, but feel free to reopen if necessary.

PR #851, opened 2024-12-29
Add docuseal

### LINKED ISSUE #847: add Docuseal (requires text/plain support or hack)
Docuseal [now exposes its version at `/version`](https://github.com/docusealco/docuseal/pull/400#issuecomment-2555514960), but the response is a `text/plain` string like `1.8.3a`, not JSON. Can Homer somehow use this as-is?

I tried creating a custom service / smart card like the others I've made, but I get `SyntaxError: JSON.parse: unexpected non-whitespace character after JSON data at line 1 column 4 of the JSON data` in the console.

I tried manually adding double quotes around the response, making it `"1.8.3a"`...

```javascript
this.versionstring = '"' + response + '"';
```

then Homer accepts the string, but it prints `Version "1.8.3a"` on the smart card (with double quotes).

meonkeys:
## Description

Add Docuseal smart card.

Fixes #847 

## Type of change

## Checklist:

- [ ] I have tested my code for new features & regressions on both mobile & desktop devices, using the latest version of major browsers. **not checking this off because I have done only minimal testing** 


src/components/services/Docuseal.vue
```diff
@@ -0,0 +1,87 @@
...
+  },
+  methods: {
+    fetchStatus: async function () {
+      this.fetch("/version", null, false)
+        .then((response) => {
+          this.status = "online";
+          this.versionstring = response;
+        })
+        .catch((e) => {
+          this.status = "offline";
```
bastienwirtz, line 50: Maybe also set the version to `unknown` or `n/a` ?
```suggestion
          this.status = "offline";
          this.versionstring = "n/a";
```

src/mixins/service.js
```diff
@@ -56,7 +56,7 @@ export default {
           throw new Error(`Ping: target not available (${response.status} error)`);
         }
 
-        return json ? response.json() : response;
+        return json ? response.json() : response.text();
```
bastienwirtz, line 59: Good catch! :pray:

bastienwirtz (approved):

src/components/services/Docuseal.vue
meonkeys: I prefer it to remain `null` if the fetch fails, since that'll show nothing for a subtitle instead of `Version n/a` or `Version unknown`, which doesn't really add value. We know it's offline anyway. What do you think?

I'm fine either way. If you think we should use `n/a` or `unknown`, then we should probably do that for the other custom services that currently leave it as `null` if the fetch fails (Wallabag, Gitea/Forgejo, Nextcloud, Olivetin).

bastienwirtz (approved):

Issue #855, opened 2025-01-06
It takes ~30s for the container to be marked as healthy unless customizing the healthchecks


tanasegabriel:
**Describe the bug**
The default healthcheck runs on a 30s interval:

https://github.com/bastienwirtz/homer/blob/d5921d939b80bcf3bccb3bcfe4f811567d1ff93a/Dockerfile#L40-L41

This implies a ~30s wait in between starting the container and it being marked as healthy:

```bash
➜  ~ time docker compose up -d homer --wait
[+] Running 1/1
 ✔ Container homer  Healthy                                                                                       30.8s
docker compose up -d homer --wait  0.41s user 0.13s system 1% cpu 31.103 total
````

`lighttpd` boots blazingly fast; while this is not problematic when for example exposing ports, the likes of `traefik` will wait for the containers to be healthy before provisioning a router + service, etc. This results in a ~30s delay between starting the `homer` container and it being accessibly through `traefik`.

A quick fix for the compose file would be to add a healthcheck attribute as following:

```yaml
    healthcheck:
      start_period: 10s
      start_interval: 1s
```

This significantly decreases the boot time, while preserving the 30s interval for healthecks once `start_period` has elapsed or the container has been marked as healthy:

```bash
➜  ~ time docker compose up -d homer --wait
[+] Running 1/1
 ✔ Container homer  Healthy                                                                                        1.9s
docker compose up -d homer --wait  0.34s user 0.07s system 18% cpu 2.199 total
```

It would be neat if these would be baked in the Dockerfile, to avoid other users experiencing this or having to define their custom healtheck.

**Expected behavior**
Homer's container should be marked as healthy relatively soon after the `lighttpd` server has started.

**Logs & errors**
N/A

**Screenshots**
N/A

**Configuration**
If applicable, copy related homer yaml configuration here.
N/A

PR #856, opened 2025-01-06
Snappier container healthchecks

tanasegabriel:
## Description
Allowing the healthceck to mark the container as healthy shortly after `lighttpd` has started, rather than always waiting ~30 seconds.

### Before:
```bash
➜  ~ time docker compose up -d homer --wait
[+] Running 1/1
 ✔ Container homer  Healthy                                                                                       30.8s
docker compose up -d homer --wait  0.41s user 0.13s system 1% cpu 31.103 total
```

### After:
```bash
➜  ~ time docker compose up -d homer --wait
[+] Running 1/1
 ✔ Container homer  Healthy                                                                                        1.9s
docker compose up -d homer --wait  0.34s user 0.07s system 18% cpu 2.199 total
```

Reference: https://docs.docker.com/reference/dockerfile/#healthcheck

### Tested with:

```
Client: Docker Engine - Community
 Version:           27.4.1
 API version:       1.47
 Go version:        go1.22.10
 Git commit:        b9d17ea
 Built:             Tue Dec 17 15:46:31 2024
 OS/Arch:           linux/arm64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.4.1
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.10
  Git commit:       c710b88
  Built:            Tue Dec 17 15:46:31 2024
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          1.7.24
  GitCommit:        88bf19b2105c8b17560993bee28a01ddc2f97182
 runc:
  Version:          1.2.2
  GitCommit:        v1.2.2-0-g7cb3632
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

Docker Compose version v2.32.1
```

## This implies a minimum compose version of v2.20.2

```
interval, timeout, start_period, and start_interval are specified as durations. Introduced in Docker Compose version 2.20.2
```

I'm unsure if this might break older Docker versions, Synology is known for not keeping on top of updates for example 😓 

Fixes # (issue)
https://github.com/bastienwirtz/homer/issues/855

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ COPY --from=build-stage --chown=${UID}:${GID} /app/dist/assets /www/default-asse
 
 USER ${UID}:${GID}
 
-HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
+HEALTHCHECK --start-period=10s --start-interval=1s --interval=30s --timeout=5s --retries=3 \
     CMD wget --no-verbose --tries=1 --spider http://127.0.0.1:${PORT}/ || exit 1
 
 EXPOSE ${PORT}


Dockerfile
```diff
@@ -37,7 +37,7 @@ COPY --from=build-stage --chown=${UID}:${GID} /app/dist/assets /www/default-asse
 
 USER ${UID}:${GID}
 
-HEALTHCHECK --interval=30s --timeout=5s --retries=3 \
+HEALTHCHECK --start-period=10s --start-interval=1s --interval=30s --timeout=5s --retries=3 \
```
tanasegabriel, line 40: An alternative here would be to drop the `start-interval` arg (this is the one that was relatively recently introduced) and simply keep `start-period` with the default interval:

```
--start-period=DURATION (default: 0s)
--start-interval=DURATION (default: 5s)
```

That might be less disruptive for older Docker versions. @bastienwirtz wdyt?

bastienwirtz: Hello, 
Indeed it's relatively recent, but it's been released more than a year ago now, and it would be nice to see the container becoming healthy quickly. I think we can try to keep it, if it's a major issue I'll reconsider it, thanks for pointing that out @tanasegabriel, I'll keep it in mind :+1:

bastienwirtz (approved):
Nice, thanks @tanasegabriel, it looks good :+1:

Issue #769, opened 2024-04-18
 Incorrect example for background-image in example config.yml file


JamieMcH:
**Describe the bug**
Example `config.yml` file has an incorrect string for `background-image:` that doesn't work for actually pointing to a local file.

**Configuration**
The correct string should be:
```yml
background-image: "/assets/your/light/bg.png"
```
Current example misses the leading '/'

DarkenLight:
I spent 30 minutes diagnosing the issue and was checking the browser console for errors. I initially thought there was a problem with the code because requests were being made to `<ip_address>/resources` instead of `<ip_address>/assets`. Thank you for pointing out the missing leading slash in the main example— I would have spent much more time troubleshooting otherwise.

PR #871, opened 2025-02-12
Update configuration.md

robsd:
Addresses issue #769 where you have to specify an absolute path i.e. /assets/logo.png rather than a relative one i.e. assets/logo.png, since the app will look in <ip_address>/resources/assets rather than <ip_address>/assets otherwise

## Description

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -21,7 +21,7 @@ copied into the config file.
 title: "App dashboard"
 subtitle: "Homer"
 # documentTitle: "Welcome" # Customize the browser tab text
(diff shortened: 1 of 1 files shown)


bastienwirtz:
Hello, 

Since the logo is used directly in the `html` in an image tag, not from the CSS (which is located in the `ressources` directory), no `/` is necessary. it's not the same as the `background-image` property, which is used as a CSS property.
Not having the `/` makes it work as relative link when Homer is served from a subfolder.
In any case, thanks for checking this :pray:

PR #887, opened 2025-03-13
Update customservices.md

jeremymeyers:
updated paperless title to reflect active version of this project and added note about custom service info not showing up with subtitle enabled


bastienwirtz (approved):
Thx @jeremymeyers !

Issue #926, opened 2025-05-05
Pihole v6 service mandates the use of an API key


tanasegabriel:
**Describe the bug**
The new Pihole service is very explicit in requiring an API key. However, Pihole can be configured to disable authentication, e.g.:

```
FTLCONF_webserver_api_password: "" # Let Authelia deal with this
```

I explicitly make use of this as I run Authelia  + a Traefik middleware that enforces 2FA on my "sensitive" services. Homer is great at playing nice with this because it can pass the proxy credentials.

The new integration complains if `apikey` is not set up.
https://github.com/bastienwirtz/homer/blob/1afa0afd00428f320fa8dd5108d5caee51e4f245/src/components/services/PiHole.vue#L119-L124

I initially thought I can work around this by setting `apikey: ""` in Homer's config file, but that still hits the authentication / session / cache code paths and feels unnecessary. Furthermore, I also run HTTPS and make use of a middleware that alters the headers to work around CORS and trying to go down the authentication logic lands me in CORS "hell" (even though that middleware works for all of my other services that have this particularity:

`www.<< domain >>.com/:1 Access to fetch at 'https://pihole.<< domain >>.com/api/auth' from origin 'https://www. << domain >> .com' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: Redirect is not allowed for a preflight request.` 

I can probably avoid that redirect though, hopefully - I'm not super familiar with the intricacies of the new Pihole.

As a side note, when any error is encountered, the card doesn't show an error icon, it simply states that pihole is disabled and that 0% of queries were blocked.

As a test, simply removing this bit of code makes the integration work in my scenario:
https://github.com/bastienwirtz/homer/blob/1afa0afd00428f320fa8dd5108d5caee51e4f245/src/components/services/PiHole.vue#L169-L172

Similarly, an early return of `true` works to validate this:

```diff
diff --git a/src/components/services/PiHole.vue b/src/components/services/PiHole.vue
index d458ba6..36bd6e8 100644
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -121,7 +121,7 @@ export default {
           "API key is required for PiHole authentication",
           "disabled",
         );
-        return false;
+        return true;
       }

       try {
```

[...]

PR #927, opened 2025-05-06
Do not mandate the usage of an API key for Pi-hole v6

tanasegabriel:
## Description
Pihole's built-in authentication can be entirely disabled, which can be desired if you use a reverse proxy + middleware for authentication.
In this case, Homer can simply send the proxy credentials and there's no need to authenticate / cache sessions / retry authentication.

Fixes # (issue)
Closes https://github.com/bastienwirtz/homer/issues/926
## Type of change

## Checklist:

### DIFF
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -116,14 +116,6 @@ export default {
       this.sessionExpiry = null;
     },
     authenticate: async function () {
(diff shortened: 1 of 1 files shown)


bastienwirtz (approved):
Indeed, authentication should be optional. Thanks @tanasegabriel !

Issue #928, opened 2025-05-06
Pihole v6 always reports a status of enabled unless the HTTP call errors out


tanasegabriel:
**Describe the bug**

If I temporarily disable blocking, the status on the card keeps on being `enabled`. This can be confusing.
If I were to temporarily stop Pihole altogether, the status becomes `disabled` and the percentage of blocked domains shows `0`.

This is discrepant to the old v5 integration. The problem is that fetching DNS status requires an additional call to `/api/dns/blocking`.

**Expected behavior**
The Pihole card should report the accurate status of `disabled` or `enabled` as per blocking being enabled or disabled. If there's an error, the status should be updated accordingly.

**Logs & errors**
Please include any usefull information:
- Errors in your browser console (`ctrl+shift+i` or `F12`)
- If applicable, your docker container logs.

**Screenshots**

**Configuration**
If applicable, copy related homer yaml configuration here.
N/A

PR #929, opened 2025-05-06
Improve Pi-hole status

tanasegabriel:
## Description
Currently, Pihole always shows a status of `enabled`, unless one of the HTTP calls involved fails. PiHole can however be temporarily disabled and the old v5 service used to reflect that.

I'm explicitly omitting also updating this, as https://github.com/bastienwirtz/homer/pull/927 removes it (to avoid merge conflicts):

https://github.com/bastienwirtz/homer/blob/30d0265d3922a1a56a60ded5def2d54577405cf2/src/components/services/PiHole.vue#L119-L123

If there's no intention of merging https://github.com/bastienwirtz/homer/pull/927, I can amend this!

Fixes # (issue)
Closes https://github.com/bastienwirtz/homer/issues/928

## Type of change

## Checklist:

### DIFF
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -230,6 +230,12 @@ export default {
   }
 
   &.disabled:before {
+    background-color: #f5a623;
+    border-color: #e59400;
+    box-shadow: 0 0 5px 1px #f5a623;
+  }
+
+  &.error:before {
     background-color: #c9404d;
     border-color: #c42c3b;
     box-shadow: 0 0 5px 1px #c9404d;
--- a/src/components/services/PiHole.vue
+++ b/src/components/services/PiHole.vue
@@ -170,16 +170,23 @@ export default {
           const authenticated = await this.authenticate();
           if (!authenticated) return;
         }
-        const response = await this.fetch(
+        const summary_response = await this.fetch(
           `api/stats/summary?sid=${encodeURIComponent(this.sessionId)}`,
         );
 
-        if (response?.queries?.percent_blocked === undefined) {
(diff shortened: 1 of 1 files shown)


src/components/services/PiHole.vue
```diff
@@ -223,6 +230,12 @@ export default {
   }
 
   &.disabled:before {
+    background-color: #f5a623;
+    border-color: #e59400;
+    box-shadow: 0 0 5px 1px #f5a623;
+  }
```
tanasegabriel, line 236: I'd argue that we should have an amber indicator for when blocking is disabled, as nothing is really broken?

image

image

image

src/components/services/PiHole.vue
```diff
@@ -170,16 +170,23 @@ export default {
           const authenticated = await this.authenticate();
           if (!authenticated) return;
         }
-        const response = await this.fetch(
+        const summary_response = await this.fetch(
           `api/stats/summary?sid=${encodeURIComponent(this.sessionId)}`,
         );
 
-        if (response?.queries?.percent_blocked === undefined) {
+        const status_response = await this.fetch(
```
bastienwirtz, line 177: Could be great to use promises instead of await to perform the request in parallel, that would make the card load more quickly.

bastienwirtz (changes requested):

src/components/services/PiHole.vue
tanasegabriel: @bastienwirtz of course, done!

bastienwirtz (approved):
Thanks @tanasegabriel !

PR #949, opened 2025-07-04
Fix prometheous inactive alerts and print version

thibaut1304:
## Description

Fixed the counting of "inactive" alerts
Previously: only firing and pending alerts were counted from /api/v1/alerts

Fix: Added a call to /api/v1/rules to iterate through groups[].rules[] and filter those with state: inactive.
Corrected display in the badge ({{ count }}) and in the text (x inactive alerts).

Added the Prometheus version (buildinfo) on desktop only.
New query to /api/v1/status/buildinfo.
Mobile detected via window.matchMedia.

Visual UX improvement
The API status is initialized to null (instead of an empty object), which prevents the badge from flashing or alerts with a default value of 0 at startup.

For your information:
Note on Prometheus Basic Authentication
Limitation of Prometheus + Basic Auth + CORS:
When using Authorization: Basic, the browser first sends an OPTIONS request (pre-flight CORS) without an authentication header. Prometheus rejects this with a 401 without CORS headers, thus blocking the actual request.
Result: It is impossible to use the Prometheus API directly from the browser with Basic Auth.

## Type of change

## Checklist:

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -536,14 +536,13 @@ See <https://docs.portainer.io/api/access#creating-an-access-token>
 
 ## Prometheus
 
-For Prometheus you need to set the type to Prometheus and provide a url.
+For Prometheus you need to set the type to Prometheus and provide a url. This will display the version on desktops.  
 
 ```yaml
 - name: "Prometheus"
-  type: Prometheus
+  type: "Prometheus"
   logo: "assets/tools/sample.png"
   url: "http://192.168.0.151/"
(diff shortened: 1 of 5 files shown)


thibaut1304:
For the commit: f53e3f6efd
The 2 rules in the wireguard group were personal tests, they do not work and therefore we can consider them as inactive!

thibaut1304:
Capture d'écran 2025-07-05 010838
Capture d'écran 2025-07-05 010517
Capture d'écran 2025-07-05 010649

PR #951, opened 2025-07-07
fix: reset service cards state on page switch by using unique keys

thibaut1304:
### The problem encountered

When I navigated between two pages of the Homer dashboard (e.g., #admin → #cams-admin), some cards still displayed the metrics from the previous page.

### Root cause

Vue was reusing<Service> components from one page to another because they had the same index-based key (:key="index" or :key="service-${groupIndex}-${index}").
As a result, the components kept their previous state between two pages.

### The fix

I modified the keys in App.vue (horizontal and vertical layout) to be truly unique by combining currentPage, group.name, and item.name:

```
:key="`${currentPage}-${group.name}-${item.name}`"
```

✅ This forces Vue to destroy and recreate the components each time a page changes.
✅ This cleanly restarts each card from scratch → no more "ghost" values.

## Type of change

## Checklist:

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -87,7 +87,7 @@
               </h2>
               <Service
                 v-for="(item, index) in group.items"
-                :key="`service-${groupIndex}-${index}`"
+                :key="`${currentPage}-${group.name}-${item.name}`"
                 :item="item"
                 :proxy="config.proxy"
                 :class="[
@@ -120,7 +120,7 @@
               </h2>
               <Service
                 v-for="(item, index) in group.items"
-                :key="index"
+                :key="`${currentPage}-${group.name}-${item.name}`"
                 :item="item"
                 :proxy="config.proxy"
                 :class="item.class || group.class"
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -86,8 +86,8 @@
                 {{ group.name }}
               </h2>
               <Service
-                v-for="(item, index) in group.items"
-                :key="`service-${groupIndex}-${index}`"
+                v-for="(item) in group.items"
+                :key="`${currentPage}-${group.name}-${item.name}`"
```
bastienwirtz, line 90: Goup and item name can contain anything like emoji and all kind of char, it might be safer to keep the index, just adding the currentPage prefix ?

bastienwirtz (changes requested):
Nice catch @thibaut1304 ! thanks.

bastienwirtz:
Good catch @thibaut1304, thanks.
Since this part of the code has been refactored, I backported your changes here https://github.com/bastienwirtz/homer/pull/984.
Hope that works for you!

Issue #976, opened 2025-08-26
smart card content does not get updated when changing pages or doing searches


khalilbtx:
**Describe the bug**
switching between pages containing smart card does not update the content of the smart cards. only refreshing the page will update the content properly. issue can be seen when using search bar also. 

**Expected behavior**
smart card reload its content every time we switch from pages or do a search

**Screenshots**

<img width="1105" height="402" alt="Image" src="https://github.com/user-attachments/assets/84e9c8de-3eb2-402f-a20e-360552cace62" />

<img width="1108" height="368" alt="Image" src="https://github.com/user-attachments/assets/bfd2a9c1-3824-4585-9a2f-ee29f60a5113" />

<img width="1044" height="371" alt="Image" src="https://github.com/user-attachments/assets/d8413045-4c82-4bda-a399-46de4a61c13f" />

Issue #958, opened 2025-07-17
Linkding Smart Card loads Incorrectly When Switching Pages


TheItsNameless:
**Describe the bug**
When a linkding smart card is placed on a secondary page which is loaded using the links section at the top, the smart card only shows a link to linkding itself, not the bookmarks.

**Expected behavior**
When clicking on the second page, homer should automatically load the bookmarks.

**Screenshots**

https://github.com/user-attachments/assets/0d46fdff-2948-475b-a66a-a723212787cb

**Configuration**

```yml
subtitle: "this is another dashboard page"

links:
  - name: "Privat"
    icon: "fas fa-house-user"
    url: ""
  - name: "Studium"
    icon: "fas fa-clipboard-list"
    url: "#studium"

services:
  ...
  - name: "Bookm - Studium"
    icon: "fas fa-bookmark" 
    items:
      - name: "Linkding"
        logo: "assets/tools/linkding.png"
        url: "..."
        type: "Linkding"
        token: "..."
        limit: 10
        query: "#bibliothek"
```

bastienwirtz:
Hello, 

Thanks for reporting this! I believe this was fixed by https://github.com/bastienwirtz/homer/commit/61d5d0b722a738a4a0fa5f790b330f1a49d0d5b2 in the latest release, could you upgrade and see if the bug is still there ? 
Thanks!

PR #1005, opened 2025-11-22
fix: ensure smart cards re-mount correctly between pages (stable keys in ServiceGroup)

brodie1600:
## Description

This PR fixes an issue where smart card content would be incorrect or stale when switching between pages.

In my case, navigating from the primary `#config` page (which has a custom smart card, let's call it "A") to a secondary page (which has three other smart cards, two of which are custom, one of which is `OpenWeather`) would cause the `OpenWeather` tile (in this instance) to render the content of custom card "A" instead. A full page refresh on the secondary page would show the correct `OpenWeather` card again.

This matches the behavior described in #976. Also possibly related to #958 .

## Root cause

`src/components/ServiceGroup.vue` was using only the loop index to key the rendered `Service` components:

```vue
<Service
  :key="`srv-${index}`"
  ...
/>
```

Because the key was just the index, Vue reused component instances across page transitions when the number of items and their positions were similar. This meant a smart card created as one type on page A could be reused as another type on page B, but keep its internal state and behavior from page A.

## Changes

I updated the `:key` used for the `Service` component in `src/components/ServiceGroup.vue` to be stable and page/group-aware:

```vue
<Service
  :key="`srv-${groupIndex}-${index}-${item.name || item.type}`"
  :item="item"
  :proxy="proxy"
  :class="item.class || group.class"
/>
```

And in the horizontal layout:

```vue
<Service
  :key="`srv-${groupIndex}-${index}-${item.name || item.type}`"
  :item="item"
  :proxy="proxy"
  :class="[
    'column',
    `is-${12 / columns}`,
    `${item.class || group.class || ''}`,
  ]"
/>
```

This prevents Vue from reusing a smart card component instance across pages
where the item type has changed.

## Type of change

- [ ] Refactoring

## Checklist:

### DIFF
--- a/src/components/ServiceGroup.vue
+++ b/src/components/ServiceGroup.vue
@@ -4,7 +4,7 @@
     <GroupHeader v-if="group.name" :group="group" class="group-title" />
     <Service
       v-for="(item, index) in group.items"
-      :key="`srv-${groupIndex}-${item.name || item.type || index}`"
+      :key="`srv-${groupIndex}-${index}-${item.name || item.type}`"
       :item="item"
       :proxy="proxy"
       :class="item.class || group.class"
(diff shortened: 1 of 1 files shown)


bastienwirtz (approved):
Looks good, thanks for this fix @brodie1600 🙏

PR #1045, opened 2026-06-15
fix: rewrite peanut smartcard status

### LINKED ISSUE #1044: bug: PeaNUT smartcard cannot show status correctly since the upstream changes
**Describe the bug**
NUT reports `ups.status` as a combination of tokens, which means values like `OL CHAG` are valid. However, homer uses swtich to match `ups.status` with specific token like `OL`, which leads to incorrect status.

According to [src/components/services](https://github.com/bastienwirtz/homer/blob/main/src/components/services/PeaNUT.vue), the code to deal `ups.status` is:
```
...
    status_text: function () {
      switch (this.ups_status) {
        case "OL":
          return "online";
        case "OB":
          return "on battery";
        case "LB":
          return "low battery";
        default:
          return "unknown";
      }
    },
    status_class: function () {
      switch (this.ups_status) {
        case "OL":
          return "online";
        case "OB": // On battery
          return "pending";
        case "LB": // Low battery
          return "offline";
        default:
          return "unknown";
      }
    },
...
```

When we meet the status string like `OL CHAG`, we will go to `default` branch and finally get `unknown` text.

**Expected behavior**
It should use String.includes rather than switch. The new code should like:
```
...
    status_text: function () {
      const status = this.ups_status;
      if (status.includes("LB")) return "low battery";
      if (status.includes("OB")) return "on battery";
      if (status.includes("OL")) return "online";
      return "unknown";
    },
    status_class: function () {
      const status = this.ups_status;
      if (status.includes("LB")) return "offline";
      if (status.includes("OB")) return "pending";
      if (status.includes("OL")) return "online";
      return "unknown";
    },
...
```

The reasons are below:
- It deals with `ups.status` value like `OL CHAG`.
- The match order represents the emergency order(`LB` -> `OB` -> `OL` -> other).
- It maintaines forward compatibility and the single token like `OL` is still valid.
- It does not cause mismatch. The available tokens are documented in the [developer-guide](https://networkupstools.org/docs/developer-guide.chunked/new-drivers.html#_status_data). More precisely, they are:
```
OL      -- On line (mains is present)
OB      -- On battery (mains is not present)
LB      -- Low battery
HB      -- High battery
RB      -- The battery needs to be replaced
CHRG    -- The battery is charging
DISCHRG -- The battery is discharging (inverter is providing load power)
BYPASS  -- UPS bypass circuit is active -- no battery protection is available
CAL     -- UPS is currently performing runtime calibration (on battery)
OFF     -- UPS is offline and is not supplying power to the load
OVER    -- UPS is overloaded
TRIM    -- UPS is trimming incoming voltage (called "buck" in some hardware)
BOOST   -- UPS is boosting incoming voltage
FSD     -- Forced Shutdown (restricted use, see the note below)
```
And we know String.includes will not mismatch except `CHAG` and `DISCHAG`, which are homer do not deal with. It should be an enhancement issue to add support to more tokens like `CHAG` and `DISCHAG`.

**Logs & errors**
Does not affect.

**Screenshots**
Does not affect.

**Configuration**
Does not affect.

absinthe4869:
## Description

[RFC 9271](https://www.rfc-editor.org/rfc/rfc9271.pdf) gives the value requirement of ups.status:

> The status of a hardware device, such as a UPS unit, is a symbolic description of the state of the unit. It consists of a space-separated list of symbols from the set {ALARM BOOST BYPASS CAL CHRG COMM DISCHRG FSD LB NOCOMM OB OFF OL OVER RB TEST TRIM}.

Rewrite lines of PeaNUT component to show ups.status correctly.

Fixes #1044 .

## Type of change

- [ ] Refactoring

## Checklist:

### DIFF
--- a/src/components/services/PeaNUT.vue
+++ b/src/components/services/PeaNUT.vue
@@ -32,28 +32,18 @@ export default {
   }),
   computed: {
     status_text: function () {
-      switch (this.ups_status) {
-        case "OL":
-          return "online";
-        case "OB":
-          return "on battery";
-        case "LB":
-          return "low battery";
-        default:
-          return "unknown";
-      }
+      const status = this.ups_status;
+      if (status.includes("LB"))     return "low battery";
+      if (status.includes("OB"))     return "on battery";
+      if (status.includes("OL"))     return "online";
+      return "unknown";
     },
     status_class: function () {
-      switch (this.ups_status) {
-        case "OL":
-          return "online";
-        case "OB": // On battery
-          return "pending";
-        case "LB": // Low battery
-          return "offline";
-        default:
-          return "unknown";
-      }
+      const status = this.ups_status;
+      if (status.includes("LB"))     return "offline";
+      if (status.includes("OB"))     return "pending";
+      if (status.includes("OL"))     return "online";
+      return "unknown";
     },
     load: function () {
       if (this.ups_load) {


bastienwirtz (approved):
Hello @absinthe4869!
Good catch ! Thanks for the fix 🙏

PR #1055, opened 2026-07-08
Adds a new `Minecraft` service type

mwhax7:
## Description

Adds a new `Minecraft` service type that displays the status of a Minecraft server using mcsrvstat.us API.

- ✅ <kbd>running</kbd> / ❌ <kbd>stopped</kbd> indicator (with green/red dot)
- 🖥️ Server software (e.g., <kbd>Paper</kbd>, <kbd>Spigot</kbd>)
- 📦 Server version (e.g., <kbd>1.21.8</kbd>)
- 👥 Players online / max

## Type of change

- [ ] Refactoring

## Checklist:

- [x] I've read & comply with the contributing guidelines.
- [x] I've checked my modifications for any breaking changes, especially in the `config.yml` file.

## Result:

> ⁉️ A working docker/implementation is already available if you'd like to test it: https://github.com/mwhax7/homer-minecraft (old homer version)

> 🚧 I know Homer has received many improvements since then, such as the new icon system and likely other enhancements as well. My implementation could certainly be improved to better match the current codebase and aesthetics, but I simply haven't taken the time to update it yet.


tanasegabriel:
FYI, see https://github.com/bastienwirtz/homer/pull/1060 for a proposed fix for the CI failures.

docs/customservices.md
```diff
@@ -443,6 +444,20 @@ The url must be the root url of Medusa application.
...
 **API Key**: The Medusa API key can be found in General configuration > Interface. It is needed to access Medusa API.
 
+## Minecraft
+
+Displays the status of a Minecraft server using mcsrvstat.us.
+
+```yaml
+- name: "Minecraft"
+  host: "127.0.0.1"
+  port: "25565"
```
bastienwirtz: Services usually defines the port within the `url`, it keeps the configuration simple and since you don't use 
```suggestion
  url: "127.0.0.1:25565"
```

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,99 @@
...
+  <Generic :item="item">
+    <template #content>
+      <p class="title is-4">{{ item.name }}</p>
+      <p class="subtitle is-6">
+        <template v-if="status === 'running'">
+          {{ software }} | v{{ version }} | {{ players.online }}/{{ players.max }} players
+        </template>
+        <template v-else>
+          Minecraft Server
+        </template>
```
bastienwirtz: It would be great to still support the subtitle option, to stay consistant with the other services
```suggestion
        <template v-if="item.subtitle">
          {{ item.subtitle }}
        </template>
        <template v-else-if="status === 'running'">
          {{ software }} | v{{ version }} | {{ players.online }}/{{ players.max }} players
        </template>
```

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,99 @@
...
+    status: "",
+    software: "",
+    version: "",
+    players: {
+      online: 0,
+      max: 0,
+    },
+  }),
+  created() {
+    this.fetchServerStatus();
```
bastienwirtz, line 45: You can easily add support for au refresh, like this:
```suggestion
    // Set up auto-update method for the scheduler
    this.autoUpdateMethod = this.fetchServerStatus;

    // Initial data fetch
    this.fetchServerStatus();
```

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,99 @@
...
+  },
+  methods: {
+    fetchServerStatus: async function () {
+      const host = this.item.host || "127.0.0.1";
+      const port = this.item.port || 25565;
+
+      window
+        .fetch(`https://api.mcsrvstat.us/2/${host}:${port}`)
```
bastienwirtz, line 53: It might be worth mentionning in the card documentation that you rely on an external service. 
Isn't it possible to fetch stats directly from the host:port ?

mwhax7: Unfortunately not. Minecraft uses its own TCP protocol instead of HTTP, so direct queries aren't possible. An intermediary service such as mcsrvstat is therefore required, unless you have another idea to integrate this kind of request?

I know some Minecraft plugins can expose an HTTP API, but this would require additional server-side configuration and wouldn't work for every server (especially vanilla or servers where plugins cannot be installed). An intermediary service remains the only generic solution to query any Minecraft server without modifications.

docs/customservices.md
```diff
@@ -443,6 +444,19 @@ The url must be the root url of Medusa application.
 
 **API Key**: The Medusa API key can be found in General configuration > Interface. It is needed to access Medusa API.
 
+## Minecraft
+
+Displays the status of a Minecraft server using mcsrvstat.us.
+
+```yaml
+- name: "Minecraft"
+  url: "127.0.0.1:25565"
```
tanasegabriel, line 453: `url` (used when clicking on the card) and `endpoint` (used for the API calls, if missing, `url` is used) are the options that are commonly used here - can we use that approach?

I seems the card itself is using `host` and `port`, which are not documented here. I think the doc here is correct, but rather we should consider adjusting the service to match that.

That would also potentially allow people to define the external service used, instead of us being prescriptive here, so something like this:


```yaml
- name: "Minecraft"
  type: "Minecraft"
  logo: "assets/tools/minecraft.png"
  url: "127.0.0.1:25565"
  endpoint: "https://api.mcsrvstat.us/3" # potentally we can default to this, if unset by the user
```

Maybe also worth adding some notes, letting users know that this relies on an external service?

Minecraft servers speak their own TCP protocol rather than HTTP, so a browser cannot query them directly. This card therefore relies on an intermediary HTTP service, defaulting to [mcsrvstat.us](https://api.mcsrvstat.us/).

> [!IMPORTANT]
> This service relies on an external service, the address in `url` is sent to a third party every time the dashboard is loaded, and the lookup happens from their network rather than yours. 

> [!NOTE]
> `software` and `version` are optional in the API response and are free form text, so they are left out of the card when the server does not report them.

.nvmrc
```diff
@@ -0,0 +1 @@
+24
```
tanasegabriel, line 1: This can be removed, if you can sync your fork, builds should not be blocked by the absence of this.

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+  <Generic :item="item">
+    <template #content>
+      <p class="title is-4">{{ item.name }}</p>
+      <p class="subtitle is-6">
+        <template v-if="item.subtitle">
+          {{ item.subtitle }}
+        </template>
+        <template v-else-if="status === 'running'">
+          {{ software }} | v{{ version }} | {{ players.online }}/{{ players.max }} players
```
tanasegabriel, line 10: The subtitle assumes both fields are always present, but they are optional in the API response and free form. Querying `mc.hypixel.net` right now returns no `software` at all, and `version` as the string `"Requires MC 1.8 / 1.21"`, so this renders as `Unknown | vRequires MC 1.8 / 1.21 | ...`. Worth building the line from whatever is actually present, and dropping the `v` prefix since `version` is not a version number.

```suggestion
          {{ details }}
```

That relies on a `details` computed, see below.

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,103 @@
...
+  },
+  data: () => ({
+    status: "",
+    software: "",
+    version: "",
+    players: {
+      online: 0,
+      max: 0,
+    },
+  }),
```
tanasegabriel, line 39: Two computed properties would help here. `server` normalises `url` into the bare `host[:port]` the status API expects, so people can write a normal URL. `details` renders only the fields the server actually reported.

```suggestion
  data: () => ({
    status: "",
    software: "",
    version: "",
    players: {
      online: 0,
      max: 0,
    },
  }),
  computed: {
    // The status API takes a bare `host[:port]`, so drop any scheme and path.
    server: function () {
      return (this.item.url || "")
        .replace(/^[a-z][a-z0-9+.-]*:\/\//i, "")
        .replace(/\/.*$/, "");
    },
    details: function () {
      const players = `${this.players.online}/${this.players.max} players`;
      return [this.software, this.version, players].filter(Boolean).join(" | ");
    },
  },
```

This also needs `const DEFAULT_API = "https://api.mcsrvstat.us/3";`

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,103 @@
...
+  created() {
+    // Set up auto-update method for the scheduler
+    this.autoUpdateMethod = this.fetchServerStatus;
+
+    // Initial data fetch
+    this.fetchServerStatus();
+  },
```
tanasegabriel, line 46: Following up on my comment on the docs: this is where the default endpoint can live. The `service.js` mixin sets `this.endpoint = this.item.endpoint || this.item.url` in its own `created`, and mixin hooks run before the component's, so overriding it here gives a sensible default that the user can replace with `endpoint`.

```suggestion
  created() {
    // Minecraft speaks its own TCP protocol, so an HTTP intermediary is required.
    // Point `endpoint` at a self hosted one to avoid the third party default.
    if (!this.item.endpoint) {
      this.endpoint = DEFAULT_API;
    }

    // Set up auto-update method for the scheduler
    this.autoUpdateMethod = this.fetchServerStatus;

    // Initial data fetch
    this.fetchServerStatus();
  },
```

src/components/services/Minecraft.vue
```diff
@@ -0,0 +1,103 @@
...
+            this.players.max = data.players?.max || 0;
+          } else {
+            this.status = "stopped";
+          }
+        })
+        .catch((e) => {
+          console.log(e);
+          this.status = "error";
+        });
+    },
```
tanasegabriel, line 70: A few things bundled together here.

`window.fetch` bypasses the `service.js` mixin, so per item `headers`, `useCredentials` and `successCodes` are all skipped. Maybe that's fine for an external service though, so I'm open for pushback here!

Also, he `127.0.0.1` fallback is worth removin. The lookup happens on the API provider's network rather than yours, so an unset host means mcsrvstat resolves `127.0.0.1` from its own side instead of reaching your server. Failing loudly is more useful than silently querying the wrong thing, I'd say?

```suggestion
    fetchServerStatus: async function () {
      if (!this.server) {
        console.error(
          `Minecraft: "${this.item.name}" is missing the url option`,
        );
        this.status = "error";
        return;
      }

      try {
        const data = await this.fetch(this.server);

        if (!data.online) {
          this.status = "stopped";
          return;
        }

        this.status = "running";
        // Both are optional and free form: plenty of servers report neither.
        this.software = data.software || "";
        this.version = data.version || "";
        this.players.online = data.players?.online || 0;
        this.players.max = data.players?.max || 0;
      } catch (e) {
        console.error(e);
        this.status = "error";
      }
    },
```

PR #1057, opened 2026-07-13
fix(UptimeKuma): stop overwriting the configured url

### LINKED ISSUE #1056: UptimeKuma overwrites `url`, forcing `/status/<slug>` on the card link and corrupting the card's URL
**Describe the bug**

`UptimeKuma.vue` rewrites its own card link in `created()`:

```js
/* eslint-disable */
this.item.url = `${this.item.url}/status/${this.dashboard}`;
```

This appears to be the sole service that does not respect the documented pattern. [`docs/customservices.md` states under Common options](https://github.com/bastienwirtz/homer/blob/4895c3a0c40083e985b7233e163ee2131e4c48ec/docs/customservices.md#common-options) that `url` is the "Card link and API base url", and every other service leaves `item.url` alone, letting `Generic.vue`'s `<a :href="item.url">` render it. `UptimeKuma` is the only one that mutates it:

```
$ grep -rn "item\.url\s*=" src/
src/components/services/UptimeKuma.vue:112:    this.item.url = `${this.item.url}/status/${this.dashboard}`;
```

That is also why the line needs `/* eslint-disable */`, to silence the Vue prop mutation rule.

Consequences:

1. The user cannot customize the card link. Whatever `url` is set to, `/status/<slug>` gets appended.
2. It bypasses Uptime Kuma's own `Settings > General > Entry Page` option, which already lets the user choose between the dashboard and any status page for anyone hitting the root URL. That setting is server side, more flexible, and applies to everyone, so Homer hardcoding a deep link is both redundant and more restrictive.
3. The mutation is not idempotent, so the link corrupts itself. `item` objects are shared references from the parsed config. `filterServices()` (`src/App.vue:250`) rebuilds the service list from those same objects, and `isVertical` (`vlayout && !filter`) flips when a filter is active, so using the search box remounts the card, runs `created()` again, and appends the path a second time.

**Expected behavior**

`url` is the card link and is never rewritten, consistent with every other service. `endpoint` overrides the API base when needed, as per Common options. `slug` only selects which status page the card reads its status and uptime from.

That would let the user either point `url` wherever they want (with `endpoint` for the API base), or simply leave it at the Uptime Kuma root and let the `Entry Page` setting decide the destination.

**Logs & errors**
N/A

**Additional context**
N/A

tanasegabriel:
## Description

Fixes https://github.com/bastienwirtz/homer/issues/1056

`UptimeKuma` is the only service that rewrites its own card link:

```js
this.item.url = `${this.item.url}/status/${this.dashboard}`;
```

This removes the mutation so the card follows the same `url` / `endpoint` contract as every other service.

**Changes**

- Remove the `this.item.url` mutation from `created()`, along with the `/* eslint-disable */` it needed to silence the Vue prop mutation rule.
- Rename the `dashboard` computed to `slug`, matching the config key it reads. Its only job is to pick which status page the API calls target, so `dashboard` read as the opposite of what it selects.

No `endpoint` handling is added, it is inherited from the `service` mixin.

**Behavior**

`url` is the card link and is never rewritten. `slug` only selects which status page the card reads its status and uptime from.

**Breaking change**

Cards previously linked to `<url>/status/<slug>` and now link to `url` as configured. Uptime Kuma's "root" is a backwards compativle replacement in the majority of cases, but under specific scenarios, it might redirect somewhere else. To keep the old URL as the destination, either:

- set Uptime Kuma's `Settings > General > Entry Page` to that status page, which is server side and applies to everyone, or
- point `url` straight at the status page, and set `endpoint` to the Uptime Kuma base url so the API calls still resolve, in line with every other service.

**Also fixed**

The mutation was not idempotent. Because `item` objects are shared references from the parsed config and `filterServices()` rebuilds the list from them, using the search box remounted the card and appended `/status/<slug>` again, giving `.../status/my-slug/status/my-slug`.

## Type of change

- [ ] Refactoring

## Checklist:

- [ ] I have made corresponding changes to the documentation (`README.md`).


tanasegabriel:
CI is broken, see https://github.com/bastienwirtz/homer/pull/1060 for a proposed fix.

bastienwirtz (approved):
Thanks @tanasegabriel!
It looks better indeed. 
It might be worth adding a note about the breaking change in the [service doc](https://github.com/bastienwirtz/homer/blob/main/docs/customservices.md#uptime-kuma)  in case it break something for someone ? Explaining how to change the config?

tanasegabriel:
>It might be worth adding a note about the breaking change

That is a good shout @bastienwirtz. I just added that, let me know if you have any suggestions for the wording there, please. 

Thanks!

bastienwirtz (approved):
Sounds Good!

Issue #971, opened 2025-08-19
Additional headers for service are not populated


alexeyklots:
**Describe the bug**

I tried to set few additional headers on the service level or proxy level of Homer configuration. 
Still, they are working only for those services that do not already have some internal headers defined. 
For example, it is working fine for the Uptimekuma service (I looked in the source code and did not find any headers configuration)

However, when I attempted to use Portainer or Proxmox services, the necessary headers were not set. In the source code, I noticed that headers are already defined, so it appears that the headers defined at the proxy level of the home configuration are not properly joined with the service-specific headers.
https://github.com/bastienwirtz/homer/blob/main/src/components/services/Portainer.vue#L91
https://github.com/bastienwirtz/homer/blob/main/src/components/services/Proxmox.vue#L116

**Expected behavior**
Custom headers should be set for services if defined

PR #1065, opened 2026-07-16
Add Seerr custom service

### LINKED ISSUE #1064: Seerr custom service

[Seerr](https://github.com/seerr-team/seerr) is a self hosted request manager for Plex, Jellyfin and Emby libraries.  It is the community successor to Overseerr and Jellyseerr, and ships migration guides for both.

Homer has no smart card for it, so a Seerr instance can only ever be a static link. Everything worth knowing at a glance (how many requests are waiting for approval, how many are currently downloading, whether users have reported problems with what already landed) means opening Seerr itself. Homer already has smart cards for the tools sitting behind Seerr (Radarr, Sonarr, Tautulli, Jellystat), but not for the request manager sitting in front of them.

A `Seerr` smart card with the following features:

- **Subtitle**: the Seerr version, with an icon when an update is available and/or when a restart is required.
- **Badges**: available media, pending requests, processing requests, open issues.
- Everything toggleable through a `monitor` tree, everything on by default, so the card can be trimmed down to just a version subtitle.

- **Leaving it as a static card.** Works today, but shows nothing beyond the link.
- **The Ping card.** Reports liveness only, which is the least interesting thing about a request manager.
- **Reusing a potential Overseerr or Jellyseerr card.** Seerr keeps the same `/api/v1` surface and `X-Api-Key` authentication as both, so it should work largely unchanged against them, but neither of these two exist as of now.

tanasegabriel:
## Description

[Seerr](https://github.com/seerr-team/seerr) is a self hosted request manager for Plex, Jellyfin and Emby libraries, and the community successor to Overseerr and Jellyseerr. Homer has no card for it, so a Seerr instance can only be a static link: anything worth knowing at a glance means opening Seerr itself.

This adds a `Seerr` card. The subtitle shows the version, with an up arrow icon when an update is available and a recycle icon when a restart is required. Four badges cover available media, pending requests, processing requests and open issues. Badges with a count of `0` are hidden, so a quiet instance stays visually quiet and only speaks up when something needs attention.

Display is controlled with a flat optional `hide` list, matching the existing Proxmox card. Everything shows by default; you only list what you want gone:

```yaml
- name: "Seerr"
  type: "Seerr"
  url: "http://seerr.example.com"
  apikey: "<---insert-api-key-here--->"
  hide:
    - media            # drop the available media badge
    - restartRequired  # drop the restart icon, keep the update one
```

Valid keys: `updateAvailable` and `restartRequired` (the two subtitle icons), plus `media`, `pending`, `processing` and `issues` (the four badges). Hide all six for a version-only card.

### Changes since review

* Replaced the nested `monitor` option with the flat `hide` list above, matching Proxmox, per the standardization point raised in review.
* Added two small reusable helpers to the `service.js` mixin so any card can share them: `isValueShown(key)` (reads `item.hide`, the same shape Proxmox already uses) and `capCount(value, max = 99)`.
* The `/status` endpoint is now part of the auto refresh cycle, so the version and its icons stay current instead of being fetched only once.
* Dropped the computeds that held static config and the `?? null` on the version string.

### Other notes

* Endpoints used, all under `/api/v1`: `/status` for the version and its two flags, `/media?filter=allavailable` for available media, `/request/count` for pending and processing, `/issue/count` for open issues.
* `apikey` is sent as an `X-Api-Key` header. It is only needed for the badges: `/status` is public, so the version subtitle works without one.
* `allavailable` counts partially available shows (some seasons in) alongside fully available ones. No `take` is sent, since `pageInfo.results` is a total regardless of page size.
* The issues badge counts `open` only. Seerr also breaks issues down by type, but that spans the same issues as the open/closed split, so counting both would count each issue twice.
* The endpoints are fetched in parallel and auto refresh runs through the existing `updateIntervalMs` scheduler. Each call keeps its own error state, so one failing endpoint leaves the rest of the card intact instead of blanking it.
* Hiding every badge behind an endpoint skips that request entirely.
* Badge counts are capped at `99+` (via the shared `capCount`), since an available media count can run into the thousands and a badge is a small pill.
* `apikey` is not sent as an empty `headers` object when unset, so any `proxy.headers` or per item `headers` survive.
* Mock data is included under `dummy-data/seerr/` for every endpoint, so `pnpm mock` exercises all four badges and both icons.
* No new dependencies.

Not breaking: the card is new, and `hide` is new and optional.

> **Shared mixin helpers**: this PR and #1063 (Traefik) both add the identical `isValueShown` / `capCount` methods to `src/mixins/service.js`. Whichever merges second just needs the duplicate block dropped; no card code changes either way.

Closes #1064

## Card with everything enabled

image

## Type of change

- [ ] Refactoring

## Checklist:

### DIFF
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -54,6 +54,7 @@ Available services are located in `src/components/`:
 - [rTorrent](#rtorrent)
 - [SABnzbd](#sabnzbd)
 - [Scrutiny](#scrutiny)
+- [Seerr](#seerr)
 - [Speedtest Tracker](#speedtesttracker)
 - [Tautulli](#tautulli)
 - [Tdarr](#tdarr)
@@ -790,6 +791,30 @@ Displays info about the total number of disk passed and failed S.M.A.R.T and scr
 ```
 
 Auto refresh is supported by this integration.  
+
+## Seerr
+
+Displays the Seerr version as the subtitle, with icons next to it when an update is available or a restart is required, plus badges counting available media, pending requests, processing requests and open issues.
+
+```yaml
+- name: "Seerr"
+  type: "Seerr"
+  logo: "assets/tools/sample.png"
+  url: "http://seerr.example.com"
+  apikey: "<---insert-api-key-here--->"
+  # subtitle: "Requests"  # (Optional) Overrides the version subtitle.
+  # hide: []              # (Optional) hides items. Possible values are
+  #                       # "updateAvailable", "restartRequired", "media",
+  #                       # "pending", "processing" and "issues".
+```
+
+Auto refresh is supported by this integration.
+
+**Authentication**: generate an API key under **Settings > General** in your Seerr instance. The badges need it; the version subtitle works without one.
+
+**Hiding items**: everything is shown by default; list a key under `hide` to remove it. `updateAvailable` and `restartRequired` are the two subtitle icons; `media`, `pending`, `processing` and `issues` are the four badges. Hide all six for a version-only card. Badge counts are capped at `99+`.
+
+**Available media**: counts partially available shows (those with some requested seasons still missing) alongside fully available ones.
 
 ## SpeedtestTracker
 
--- a/src/components/services/Seerr.vue
+++ b/src/components/services/Seerr.vue
@@ -1,0 +1,192 @@
+<template>
+  <Generic :item="item">
+    <template #content>
+      <p class="title is-4">{{ item.name }}</p>
+      <p class="subtitle is-6">
+        <template v-if="item.subtitle">{{ item.subtitle }}</template>
+        <template v-else-if="versionstring">
+          Version {{ versionstring }}
+          <i
+            v-for="icon in visibleIcons"
+            :key="icon.key"
+            class="state"
+            :class="icon.class"
+          ></i>
+        </template>
+      </p>
+    </template>
+    <template #indicator>
+      <div class="notifs">
+        <strong
+          v-for="badge in visibleBadges"
+          :key="badge.key"
+          class="notif"
+          :style="{ backgroundColor: badge.color }"
+          :title="badge.label"
+        >
+          {{ badge.text }}
+        </strong>
+        <strong
+          v-if="serverError"
+          class="notif errors"
+          title="Connection error to Seerr API, check url and apikey in config.yml"
+        >
+          ?
+        </strong>
+      </div>
+    </template>
+  </Generic>
+</template>
+
+<script>
+import service from "@/mixins/service.js";
+
+// Key doubles as the data key its response lands on.
+const endpoints = {
+  status: "/api/v1/status",
+  // allavailable counts partially available too; results is a total, so take is moot.
+  media: "/api/v1/media?filter=allavailable",
+  requests: "/api/v1/request/count",
+  issues: "/api/v1/issue/count",
+};
+
+// Subtitle icons; key is also the status flag each one reads.
+const icons = [
+  { key: "updateAvailable", class: "fa-solid fa-circle-arrow-up" },
+  { key: "restartRequired", class: "fa-solid fa-recycle" },
+];
+
+// Indicator badges; count reads the response landed by `from`.
+const badges = [
+  {
+    key: "media",
+    label: "Available media",
+    color: "#2ac194",
+    from: "media",
+    count: (data) => data?.pageInfo?.results,
+  },
+  {
+    key: "pending",
+    label: "Pending requests",
+    color: "#0a4bc4",
+    from: "requests",
+    count: (data) => data?.pending,
+  },
+  {
+    key: "processing",
+    label: "Processing requests",
+    color: "#6f11a6",
+    from: "requests",
+    count: (data) => data?.processing,
+  },
+  {
+    key: "issues",
+    label: "Open issues",
+    color: "#c1942a",
+    from: "issues",
+    count: (data) => data?.open,
+  },
+];
+
+export default {
+  name: "Seerr",
+  mixins: [service],
+  props: {
+    item: Object,
+  },
+  data: () => ({
+    status: null,
+    requests: null,
+    issues: null,
+    media: null,
+    failed: {},
+  }),
+  computed: {
+    serverError() {
+      return Object.values(this.failed).some(Boolean);
+    },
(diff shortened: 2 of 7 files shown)


src/components/services/Seerr.vue
```diff
@@ -0,0 +1,219 @@
...
+      );
+    },
+  },
+  created() {
+    if (this.monitoredEndpoints.length) {
+      this.autoUpdateMethod = this.fetchStats;
+      this.fetchStats();
+    }
+    // A subtitle override hides the version, so don't fetch it.
+    if (!this.item.subtitle) this.load("status");
```
bastienwirtz: Just to make sure is was intentional, this won't be included in the auto refresh system, is that ok for you?

src/components/services/Seerr.vue
```diff
@@ -0,0 +1,219 @@
...
+  },
+  data: () => ({
+    status: null,
+    requests: null,
+    issues: null,
+    media: null,
+    failed: {},
+  }),
+  computed: {
+    fetchOptions() {
```
bastienwirtz: This does not have to be a computed property. It's meant to be used for values calculated from other reactive data. they are reactive functions that automatically re-run when their dependencies change, but here nothing will change, nothing have to be watched.

src/components/services/Seerr.vue
```diff
@@ -0,0 +1,219 @@
...
+      if (!this.item.apikey) return {};
+      return { headers: { "X-Api-Key": this.item.apikey } };
+    },
+    serverError() {
+      return Object.values(this.failed).some(Boolean);
+    },
+    monitored() {
```
bastienwirtz: Same here I think, since the config and monitor configuration is static.

src/components/services/Seerr.vue
```diff
@@ -0,0 +1,219 @@
...
+        .map((badge) => {
+          const value = this.monitored[badge.key]
+            ? badge.count(this[badge.from]) || 0
+            : 0;
+          return { ...badge, value, text: cap(value) };
+        })
+        .filter((badge) => badge.value > 0);
+    },
+    versionstring() {
+      return this.status?.version ?? null;
```
bastienwirtz: nit: I don't think the "?? null" part is necessary, but it don't hurt

src/components/services/Seerr.vue
```diff
@@ -0,0 +1,219 @@
...
+// A section nests one level deeper: false, true or a map.
+function monitorsIn(monitor, section, key) {
+  if (!monitors(monitor, section)) return false;
+  return monitors(monitor[section] ?? true, key);
+}
+
+// A badge is a small pill; large numbers blow out the card.
+function cap(value) {
+  return value > 99 ? "99+" : value;
+}
```
bastienwirtz: Might be nice to make this globally available (in the mixin maybe?) to all services it can be a consistent thing across all cards

docs/customservices.md
```diff
@@ -791,6 +792,34 @@ Displays info about the total number of disk passed and failed S.M.A.R.T and scr
...
+  url: "http://seerr.example.com"
+  apikey: "<---insert-api-key-here--->"
+  # subtitle: "Requests"   # (Optional) Overrides the version subtitle.
+  # monitor:               # (Optional) Hide things; anything you don't mention stays on.
+  #   media: false             # hide the available media badge
+  #   issues: false            # hide the open issues badge
+  #   server: false            # hide both subtitle icons
+  #   requests:                # or name them individually to hide only part of a group
+  #     pending: false             # hide the pending badge, keep the processing one
+  # monitor: false         # (Optional) Or hide every icon and badge (version-only card).
```
bastienwirtz: I think all card could use an option to choose what to display, we may have an opportunity to define a standard way to do it. [Proxmox](https://github.com/bastienwirtz/homer/blob/main/docs/customservices.md#proxmox) use a `hide` key, which is explicit and simple, that might be a good way to do it everywhere?

bastienwirtz (commented):
Hey @tanasegabriel thanks for this integration! 

It looks great, my only concern is that I find the `monitor` option a but complex for what it is, we might be able to simplify it and makes it a common thing (cf my comments on the md file)

What do you think?

src/components/services/Seerr.vue
```diff
@@ -0,0 +1,192 @@
...
+    visibleIcons() {
+      return icons.filter(
+        (icon) => this.isValueShown(icon.key) && this.status?.[icon.key] === true,
+      );
+    },
+  },
+  created() {
+    this.fetchOptions = this.item.apikey
+      ? { headers: { "X-Api-Key": this.item.apikey } }
+      : {};
```
tanasegabriel, line 130: ### Follow-up proposal: header handling in the `service.js` mixin

Flagging an existing sharp edge I ran into here, as a possible follow-up (deliberately kept out of this PR to keep it focused).

In `src/mixins/service.js`, `fetch` resolves `proxy.headers` / per-item `headers` into `options.headers` and then does:

```js
options = Object.assign(options, init);
```

That is a shallow assign, so when a card passes `{ headers: { ...auth } }` as `init`, the card's headers object **replaces** `options.headers` wholesale, dropping any `proxy.headers` or per-item `headers` the user configured. This affects every service card (report in https://github.com/bastienwirtz/homer/issues/971). The `apikey`/`basic_auth` "not sent as an empty `headers` object when unset" guard in this PR only covers the no-auth case; the auth-set case still clobbers.

A small mixin-level change would fix it for all cards at once, by merging instead of replacing:

```js
if (init?.headers) {
  init = { ...init, headers: { ...options.headers, ...init.headers } };
}
options = Object.assign(options, init);
```

Auth headers still win on key conflicts (same as today); the difference is that non-conflicting `proxy.headers` / item `headers` now survive. Happy to open a separate PR for this if it sounds reasonable.

src/components/services/Seerr.vue
tanasegabriel: Not intensional, an oversight. Thanks for spotting it!

docs/customservices.md
tanasegabriel: That's way neater, imlpemented! There might be further room for QoL changes like this, good idea!

src/components/services/Seerr.vue
tanasegabriel: You're right, moved to static, thanks!

bastienwirtz (approved):

PR #1074, opened 2026-08-07
fix: merge custom headers with the ones set by services

### LINKED ISSUE #971: Additional headers for service are not populated

### LINKED ISSUE #1008: Prometheus service type doesn't apply basic_auth credentials
## Description

The Prometheus service type doesn't include the Authorization header when making API requests, even when `basic_auth` is configured. This differs from the Traefik service type which correctly applies `basic_auth`.

## Homer version

v25.11.1

## Configuration

```yaml
- name: "Prometheus"
  type: "Prometheus"
  url: "https://prometheus.example.com"
  basic_auth: "admin:password"

- name: "Traefik Dashboard"
  type: "Traefik"
  url: "https://traefik.example.com"
  basic_auth: "admin:password"
```

## Observed behavior

Network requests show:
- **Traefik**: Sends OPTIONS preflight → GET with Authorization header → ✅ 200 OK
- **Prometheus**: Sends GET without Authorization header → ❌ 401 Unauthorized

## Expected behavior

Both service types should include the Authorization header when `basic_auth` is configured.

## Additional context

- Server correctly returns CORS headers on 401 responses (`Access-Control-Allow-Origin`, `Access-Control-Allow-Credentials: true`)
- This is NOT the CORS preflight issue documented in PR #949 - the problem is that the Authorization header is never sent at all for the Prometheus type
- The Traefik component correctly triggers a CORS preflight by including the Authorization header

tanasegabriel:
## Description

Custom headers set through `proxy.headers` or a service's own `headers` key are silently dropped by any card that sends headers of its own, which is what [#971](https://github.com/bastienwirtz/homer/issues/971) reports for Portainer and Proxmox. 23 of the 53 service cards send headers, across 34 call sites, so this affects a large part of the smart card surface.

The cause is in the `service.js` mixin. `this.fetch()` assigned the configured headers to the request options, then merged the card's `init` over them with `Object.assign`. Because `headers` is a single key, `init.headers` replaced the whole configured set instead of adding to it, so a card sending `X-Api-Key` erased the user's `Authorization` header rather than sending both.

The mixin now merges the three sources in order of specificity:

```js
options.headers = mergeHeaders(
  this.proxy?.headers,
  this.item.headers,
  init?.headers,
);
```

Precedence is `proxy.headers`, then the item `headers`, then the ones the card builds itself, each layer overriding the previous one by name. Two details were needed to make that correct:

**Header names are compared case-insensitively.** HTTP header names are case-insensitive but object keys are not, so merging `{authorization: ...}` from the configuration with `{Authorization: ...}` from a card would keep both keys, and `fetch` would join them into a single `authorization: Basic aaa, Bearer bbb` value that servers reject. Names are lowercased while merging, so one header name always means one header.

**Unset values no longer count as values.** Several cards build their headers unconditionally, for example Portainer with `{"X-Api-Key": this.item.apikey}`. When that config key is absent the value is `undefined` and `fetch` sends the literal string `"undefined"`, overwriting a header the user did set. Undefined and null values are skipped, so the configured header survives.

**Cards that only sometimes send a header are fixed too.** Traefik, TrueNAS Scale and HyperHDR build `let headers = {}` and fill it only when their optional auth key is present. That empty object used to erase the configured headers on every request where the key was unset. Same bug as #971, never reported separately.

**Plex** was the last card still using the native `fetch` for part of its work, so its library requests bypassed the proxy configuration entirely. Its two remaining native calls now go through `this.fetch`, which also gives it `successCodes` support.

### Verification

The repository has no test suite, so the change was checked by diffing the requests the mixin builds before and after, rather than by inspection:

* Every `this.fetch()` call in `src/components/services` was parsed to collect the shape of its `init` argument: 88 call sites in eight distinct shapes, 34 of them carrying the card's own headers.
* Those eight shapes, plus the two "optional config key unset" variants, were run against seven proxy and item configurations, comparing the complete request options with header lists normalised through `new Headers()` so that casing and key order cannot hide a difference. 44 of the 70 combinations are byte identical.
* **For a dashboard that sets neither `proxy.headers` nor item `headers`, which is the default, only one of the ten shapes changes at all**: a card header whose value is `undefined` is no longer sent. Every other call site builds exactly the request it built before.
* The remaining 25 differences are configured headers now reaching cards that used to discard them, which is the fix itself.
* Nothing outside the header path moved. `method`, `cache`, `body`, `credentials` and the `AbortSignal` the Ping card passes are identical in every combination.

`pnpm lint` passes and a production build succeeds.

### Other notes

* Item `headers` used to replace `proxy.headers` entirely. They now layer, so a header set only at the proxy level survives alongside item headers. `docs/customservices.md` documents the new precedence.
* Plex now inherits `proxy.useCredentials`. A dashboard that enables it globally will send credentialed requests to Plex, which answers `Access-Control-Allow-Origin: *` and therefore refuses them. `useCredentials: false` on the item restores the previous behaviour. Plex also now shows a connection error instead of zero counts when the server answers with an error status.
* `docs/development.md` gains a short "Fetching data" section and `AGENTS.md` the matching rules, so a new card does not reintroduce the bug by reading `proxy.headers` itself.
* This does not close [#1008](https://github.com/bastienwirtz/homer/issues/1008). The Prometheus card sends no headers at all and does not read `basic_auth`, which needs its own change. Setting `headers: {Authorization: "Basic <base64>"}` on the item works today.
* No new dependencies and no lockfile change.

Not breaking: a dashboard that sets neither `proxy.headers` nor item `headers` sends exactly the same requests as before.

Closes #971

## Type of change

- [ ] Refactoring

## Checklist:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,8 @@ All service components follow this architecture:
 - Extend `Generic.vue` using Vue slots (`<template #indicator>`, `<template #content>`, `<template #icon>`)
 - Use the `service.js` mixin (`src/mixins/service.js`) for common API functionality
 - Use a custom `fetch` method provided by the service mixin to seamlessly support proxy configuration, custom headers, and credentials.
+- Never call the native `fetch` in a service component. `OpenWeather` (third-party public API) and `Rtorrent` (XML-RPC on a separate host) are the only exceptions, and are not examples to follow.
+- Pass the service's own headers (an API key, an `Authorization` header, ...) as the `init.headers` argument of `this.fetch(path, init, json)`. They layer over the user's `proxy.headers` and item `headers` rather than replacing them, so a card must never build its headers from `proxy` or `item.headers` itself. Names are compared case-insensitively, and the card value wins on conflict.
 
 ### Auto-Update Configuration
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,7 +31,7 @@ All service components follow this architecture:
 - Extend `Generic.vue` using Vue slots (`<template #indicator>`, `<template #content>`, `<template #icon>`)
 - Use the `service.js` mixin (`src/mixins/service.js`) for common API functionality
 - Use a custom `fetch` method provided by the service mixin to seamlessly support proxy configuration, custom headers, and credentials.
-- Never call the native `fetch` in a service component. `OpenWeather` (third-party public API) and `Rtorrent` (XML-RPC on a separate host) are the only exceptions, and are not examples to follow.
+- Never call the native `fetch` in a service component unless you have a specific reason not to. See `OpenWeather` (third-party public API) or `Rtorrent` (XML-RPC on a separate host) for use case where the service fetch is not suitable.
 - Pass the service's own headers (an API key, an `Authorization` header, ...) as the `init.headers` argument of `this.fetch(path, init, json)`. They layer over the user's `proxy.headers` and item `headers` rather than replacing them, so a card must never build its headers from `proxy` or `item.headers` itself. Names are compared case-insensitively, and the card value wins on conflict.
 
 ### Auto-Update Configuration
--- a/docs/development.md
+++ b/docs/development.md
@@ -29,7 +29,7 @@ Any service consuming an API must use the [`service`](https://github.com/bastien
 - **Refresh**: assign the method to re-run to `this.autoUpdateMethod`, the mixin schedules it using `updateIntervalMs`.
 
 > [!NOTE]
-> `OpenWeather` and `Rtorrent` bypass the mixin, respectively for a third-party public API and an XML-RPC host. Don't use them as an example for a new service.
+> Some services like `OpenWeather` and `Rtorrent` bypass the mixin, respectively for a third-party public API and an XML-RPC host. Don't use them as an example for a new service.
 
 ### Skeleton
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -18,6 +18,19 @@ Each one is **optional**, and will display the usual information if omitted.
 
 Each service must implement the `item` [property](https://vuejs.org/v2/guide/components-props.html) and bind it the Generic component if used.
 
+### Fetching data
+
+Any service consuming an API must use the [`service`](https://github.com/bastienwirtz/homer/blob/main/src/mixins/service.js) mixin and its `this.fetch(path, init, json)` helper instead of the native `fetch`. The mixin takes care of everything the user can configure:
+
+- **URL**: `path` is appended to the service endpoint (`item.endpoint`, or `item.url`), so it only needs the part that follows it.
+- **Headers**: pass the service ones (an API key, an `Authorization` header, ...) as `init.headers`. The mixin layers them over the user's `proxy.headers` and item `headers`, so never read those yourself. If both set the same header, the service value wins.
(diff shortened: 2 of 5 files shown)


docs/development.md
```diff
@@ -18,6 +18,19 @@ Each one is **optional**, and will display the usual information if omitted.
...
+Any service consuming an API must use the [`service`](https://github.com/bastienwirtz/homer/blob/main/src/mixins/service.js) mixin and its `this.fetch(path, init, json)` helper instead of the native `fetch`. The mixin takes care of everything the user can configure:
+
+- **URL**: `path` is appended to the service endpoint (`item.endpoint`, or `item.url`), so it only needs the part that follows it.
+- **Headers**: pass the service ones (an API key, an `Authorization` header, ...) as `init.headers`. The mixin layers them over the user's `proxy.headers` and item `headers`, so never read those yourself. If both set the same header, the service value wins.
+- **Credentials**: `proxy.useCredentials` and its per-item override are applied for you.
+- **Response**: returned as parsed JSON, or as text when `json` is `false`.
+- **Refresh**: assign the method to re-run to `this.autoUpdateMethod`, the mixin schedules it using `updateIntervalMs`.
+
+> [!NOTE]
+> `OpenWeather` and `Rtorrent` bypass the mixin, respectively for a third-party public API and an XML-RPC host. Don't use them as an example for a new service.
```
bastienwirtz:
```suggestion
> Some services like `OpenWeather` and `Rtorrent` bypass the mixin, respectively for a third-party public API and an XML-RPC host. Don't use them as an example for a new service.
```
Just in case we others are added.

AGENTS.md
```diff
@@ -31,6 +31,8 @@ All service components follow this architecture:
 - Extend `Generic.vue` using Vue slots (`<template #indicator>`, `<template #content>`, `<template #icon>`)
 - Use the `service.js` mixin (`src/mixins/service.js`) for common API functionality
 - Use a custom `fetch` method provided by the service mixin to seamlessly support proxy configuration, custom headers, and credentials.
+- Never call the native `fetch` in a service component. `OpenWeather` (third-party public API) and `Rtorrent` (XML-RPC on a separate host) are the only exceptions, and are not examples to follow.
```
bastienwirtz:
```suggestion
- Never call the native `fetch` in a service component unless you have a specific reason not to. See `OpenWeather` (third-party public API) or `Rtorrent` (XML-RPC on a separate host) for use case where the service fetch is not suitable.
```
Just to keep it more generic and always up to date. If you are ok with the idea, feel free to adjust my suggestion, i'm not a native english speaker as you know 😅

bastienwirtz (approved):
Minor idea regarding doc, otherwise it looks all good 👍 
Thanks!

bastienwirtz (approved):

Issue #1025, opened 2026-03-01
Compact service cards or sub-links (pills)


Flavien06:
Hello I am trying to optimize my dashboard for high density. I have a few issues/requests regarding the display of service items:

- Sub-links (Pills) rendering: I am trying to replicate the "Dashboard icons" or "Selfhosted community" look from the official demo using the links property within an item, but they do not render at all in my setup (default theme, latest Docker).

- Compact Cards: Is there a built-in way to make service cards smaller (half-height or mini-cards) to fit more on one screen.

- Compact Link List: A "list-only" type for items would be great to display many bookmarks in a single card without large icons or padding.

PR #1080, opened 2026-09-04
feat(cards): one shared layout for every service card

### LINKED ISSUE #997: Can't access `quick` url when it overlapped by tag and card have url defined
**Describe the bug**
Can't access `quick` url when it overlapped with tag and card have url defined (tag "point" to card url). If card dont have defined url then `quick` link overlapped with tag accessible (tag "point" to `quick` link url).

**Expected behavior**

**Logs & errors**
Please include any usefull information:
- Errors in your browser console (`ctrl+shift+i` or `F12`)
- If applicable, your docker container logs.

**Screenshots**

Image

Image

**Configuration**
If applicable, copy related homer yaml configuration here.
```yml

```

### LINKED ISSUE #1077: Tags jitter when hovering over bottom edge of card
### Description
When hovering the mouse over a tag (or any interactive element) located at the bottom edge of a card in the Homer theme, the tag continuously flickers / jitters. The hover state toggles rapidly, making it impossible to interact with the tag stably.

### Steps to Reproduce
1. Open any Homer dashboard page that displays cards with tags (e.g., services, bookmarks).  
2. Move the mouse cursor over a tag that is positioned near the lower edge of a card.  
3. Keep the mouse still – observe that the tag’s hover effect (e.g., colour change, scale, or movement) starts to shake or bounce repeatedly.

### Expected Behavior
The hover effect should be stable; the tag should not flicker or move while the mouse remains over it.

### Actual Behavior
The tag jitters constantly, with the hover effect being applied and removed in a rapid loop.

### Impact
This degrades user experience and can prevent clicking on tags near the bottom of cards.

### Additional Context
The issue appears to be caused by CSS hover styles that alter layout-affecting properties (such as `margin`, `padding`, or `border-width`), which shift the element’s position and cause the mouse to leave and re‑enter the hover area.

### Environment
- Homer version: 26.04.2
- Browser: Firefox 121] 
- OS: Windows 10

tanasegabriel:
First of two stacked PRs. This one is mainly presentation: [`Generic.vue`](https://github.com/bastienwirtz/homer/blob/daa017d/src/components/services/Generic.vue) takes over the card layout and the styling moves into [`src/assets/components/`](https://github.com/bastienwirtz/homer/tree/daa017d/src/assets/components). The old slots keep every existing card rendering, with the one exception noted below. The second PR migrates the cards onto the new props and the shared service helpers.

## What changes

- The card becomes a grid with a fixed set of zones: icon, title, subtitle, badges, status chip, quicklinks and tag. A card no longer positions anything itself.
- **Quicklinks get their own lane** at the bottom, with a different position. They previously lived inside `Generic`'s default `#content` slot, so any card overriding that slot silently dropped them; in the lane they work on every card.
- **The card link becomes a single stretched overlay** (`.card-link`) instead of an anchor wrapping the content. The HTML parser closes the outer `<a>` at the first quicklink, which is what made a `quick` url unreachable when the card had its own url.
- **Only quicklinks claim the lane row.** A card with pills gives them a row of their own and the title and subtitle sit above it, so the text and the pills stay centred on the card as a group. A tag on its own stays pinned to the card's bottom edge as it is on `main`, claiming no row, so the text stays centred and nothing moves. When the two appear together the tag unfurls into the space left beside the pills, and hovering the tag itself expands it over them.
- Nothing that overhangs the card takes the pointer while the card is idle, which removes the hover feedback loop at the card edges.
- **An unfurling tag is clipped at the card edge** instead of running into the next column, and the text ellipsises where it is cut. The lane reaches both card edges to do the clipping, with a matching padding holding the quicklinks where they were.
- Card geometry (`--card-height`, `--card-pad`, `--card-radius`, `--card-lift`) becomes CSS custom properties, overridable from an additional stylesheet. The status and badge colours become theme tokens, settable through the existing `colors:` config.
- `icon` now takes precedence over `logo` when both are set, matching group headers and what [the configuration docs](https://github.com/bastienwirtz/homer/blob/daa017d/docs/customservices.md#common-options) already promise. They previously rendered side by side.
- **[`OpenWeather`](https://github.com/bastienwirtz/homer/blob/daa017d/src/components/services/OpenWeather.vue) moves onto `Generic`.** It was the only card still building its own `.card` markup, so it was also the only one the new layout could not reach. It is a card change in an otherwise presentational PR, but leaving it behind would have meant shipping a card with no hover lift and a tag that never unfurls. Going through `Generic` also makes it honour `quick:` and a configured `subtitle:`, both of which it silently dropped before, the second against [what the docs already promise](https://github.com/bastienwirtz/homer/blob/daa017d/docs/customservices.md#L90).
- `Generic` gains `subtitle`, `status` and `badges` props plus `icon` / `subtitle` / `aside` / `badges` slots. Only `OpenWeather` uses them here; the second PR migrates the rest.
- New `src/utils/format.js` holds `capCount`, which caps a badge count at `99+`. It is a pure function rather than a mixin method, so `Generic` can use it without pulling in the service mixin it has no other use for. The mixin still exposes it for cards that have not migrated.

Fixes #1077. Fixes #997.

Also addresses part of #1025: the pills that "do not render at all" are the quicklinks bug above. Not closing it, since neither the compact cards nor the list-only item type it also asks for are in scope.

## Compatibility

`Generic` still accepts the old `#content` and `#indicator` slots, so every card that uses them keeps rendering. Everything held open for those cards carries a `TODO(cards-migrate-to-api)` marker and goes in the [follow-up](https://github.com/bastienwirtz/homer/pull/1081):

- the `#content` and `#indicator` slots themselves
- `--status-color: transparent` and the `ready` / `error` / `pending` / `in-progress` aliases in `status.scss`, while cards still emit their own state names
- the dot spacing as a `margin-inline-end` on `::before` rather than a `gap` on `.status`, so the scoped `margin-right` those cards ship overrides it instead of stacking with it
- `z-index` on `.card-aside`

One thing does change before the second PR lands: **the status chip is not clickable.** `.card-aside` has to sit above the `.card-link` overlay so unmigrated indicators keep their `title` tooltips, which are the only labels on the notification counts, and that costs the click through it. On `main` the chip opened the service. The follow-up drops the `z-index` once those counts become badges, which keep their tooltips without blocking the click.

Two consequences of the `.card-link` overlay are permanent, and worth a decision:

- Card text is no longer selectable.
- Native `title` tooltips on elements under the overlay no longer appear. This matters for `Glances`, whose `:title` is the only thing naming its bare stat icons, and for the weather description on `OpenWeather`'s icon. `.card-body [title] { position: relative; z-index: 2 }` brings the body ones back, at the cost of those spans no longer forwarding the card click.

### For users with a custom stylesheet

The card markup changed, so some selectors stop matching. The affected ones are listed in [troubleshooting](https://github.com/bastienwirtz/homer/blob/daa017d/docs/troubleshooting.md). `.card:hover` still fires, so hover rules keyed on the card itself keep working.

Tags with a `tagstyle: is-*` class also pick up Bulma's `a.tag:hover` shade now that the tag is an anchor; tags without one are unaffected.

## Please merge both PRs in the same release

On its own this looks half-migrated: the new chrome with cards still using their own status words, colours and pill placement. The split exists so the CSS can be reviewed on its own, not so the two can ship separately. Merging this one alone publishes nothing in any case, since `dockerhub.yml` and `release.yml` both trigger on `v*` tags rather than on `main`.

## Testing

- `pnpm lint` and `pnpm build` clean.
- Rendered against a config covering every smart card plus the chrome cases: quicklinks, long tags, tag-only, no tag, no subtitle, no icon, logo instead of icon, both logo and icon, and long titles that have to ellipsis.
- Checked the states cards still emit themselves against the aliases: `ready`, `error` and `pending` keep their colour and `in-progress` keeps its pulse, so no card's dot changes.
- Checked side by side with `main` in both light and dark, in the default grid layout and `layout: columns`, and across the classic, walkxcode and neon themes.
- Checked the community themes listed in [the configuration docs](https://github.com/bastienwirtz/homer/blob/daa017d/docs/configuration.md) against `main` and this branch. They all still look right, with two exceptions that already behave the same way on `main`: the Dracula theme does not apply at all, and Homer Theme v2 is wonky. Both look like abandonware rather than anything this branch changes.
- Diffed the compiled CSS against `main`. Outside the card, two like-for-like moves: `.group-logo` picks up the `margin-inline-end` the group header used to get from `.media-left`, and `main`'s `.card:hover` rule is gone, which restores the error tint on a hovered `.component-error` card. The new custom properties are the only other additions, and no existing rule reads them.

One thing left alone on purpose: `--card-shadow` is documented as the card shadow but a later rule in `base.scss` has always overridden it, on `main` too. Letting it through would deepen card shadows in every dark theme, so it belongs in its own, separate PR.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -68,3 +68,20 @@ Only one icon is lost: **`fa-vector-square` no longer exists**. Font Awesome ren
 
 > [!NOTE]
 > Font Awesome 7 also gives every icon the same width by default. Homer keeps the previous behaviour, so nothing moves after upgrading. See [icon width](configuration.md#icon-width) if you want the new look.
+
+## My custom stylesheet stopped styling the service cards
+
+Every card moved onto a shared layout, so a few selectors no longer match:
+
+- `.media`, `.media-left`, `.media-content` and `.media.no-subtitle` are gone. The icon is now `.card-icon`, and the title and subtitle sit in `.card-body`.
+- Paths through the card gained a level: `.card-wrapper` sits between the item and `.card`, and the card link is `.card > .card-content > a.card-link`.
+- `.tag` collapses with `max-width` rather than `width`, so a rule setting `width` has no effect.
+- `.quicklinks` is a flex lane at the bottom of the card, so `float` no longer applies.
+- `.card-lane` reaches both card edges and clips what overflows them, so anything positioned inside it stops at the card. The tag sits at the lane's inline end rather than overhanging by its own offset.
+- The status dot animation is named `status-pulse`.
+
+`.card:hover` still matches, so hover rules keyed on the card itself keep working.
+
+## I cannot select card text, and some tooltips stopped appearing
+
+The whole card is one link, drawn as a transparent overlay on top of the content. That overlay takes the pointer, so text inside a card can no longer be selected, and a native `title` tooltip on an element in the card body no longer opens. Tooltips on the badges and the tag are unaffected, as those sit above the overlay.
--- a/.schema/config-schema.json
+++ b/.schema/config-schema.json
@@ -61,6 +61,42 @@
         },
         "background-image": {
           "type": "string"
+        },
+        "status-online": {
+          "type": "string"
+        },
+        "status-offline": {
+          "type": "string"
+        },
+        "status-warning": {
+          "type": "string"
+        },
+        "status-busy": {
+          "type": "string"
+        },
+        "status-unknown": {
+          "type": "string"
+        },
+        "badge-info": {
+          "type": "string"
+        },
+        "badge-success": {
+          "type": "string"
+        },
+        "badge-warning": {
+          "type": "string"
+        },
+        "badge-danger": {
+          "type": "string"
+        },
+        "badge-accent": {
+          "type": "string"
+        },
+        "badge-neutral": {
+          "type": "string"
+        },
+        "badge-text": {
+          "type": "string"
         }
       }
     },
@@ -265,9 +301,53 @@
         "type": {
           "type": "string",
           "description": "Optional, loads a specific component that provides extra features. MUST MATCH a file name (without file extension) available in `src/components/services`"
+        },
+        "class": {
+          "type": "string",
+          "description": "Optional css class to add on the item, useful with a custom stylesheet. Example 'highlight-purple'"
+        },
+        "background": {
+          "type": "string",
+          "description": "Optional color for the card, to set it directly without a custom stylesheet"
+        },
+        "quick": {
+          "type": "array",
+          "description": "Shortcut links shown as small pills along the bottom of the card",
+          "items": {
+            "$ref": "#/definitions/QuickLink"
+          }
         }
       },
       "title": "Item"
+    },
+    "QuickLink": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "url": {
+          "type": "string",
+          "description": "Url of this shortcut"
+        },
+        "icon": {
+          "type": "string",
+          "description": "Fontawesome icon shown inside the pill"
+        },
+        "target": {
+          "type": "string",
+          "description": "html tag target attribute like _blank for a new page"
+        },
+        "color": {
+          "type": "string",
+          "description": "Optional background color for the pill"
+        }
+      },
+      "required": [
+        "url"
+      ],
+      "title": "QuickLink"
     }
   },
   "properties": {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -141,13 +141,20 @@ services:
     items:
       - name: "Awesome app"
         logo: "assets/tools/sample.png"
-        # Alternatively a fa icon can be provided:
+        # Alternatively a fa icon can be provided. Note that icon take precedence if both icon and logo are set.
         # icon: "fab fa-jenkins"
         subtitle: "Bookmark example"
         tag: "app"
         keywords: "self hosted reddit" # optional keyword used for searching purpose
         url: "https://www.reddit.com/r/selfhosted/"
         target: "_blank" # optional html tag target attribute
+        # Optional shortcut links, shown as small pills along the bottom of the card.
+        quick:
+          - name: "Wiki"
+            url: "https://www.reddit.com/r/selfhosted/wiki/"
+            icon: "fas fa-book" # optional
+            target: "_blank" # optional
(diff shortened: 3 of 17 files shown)


docs/troubleshooting.md
```diff
@@ -68,3 +68,20 @@ Only one icon is lost: **`fa-vector-square` no longer exists**. Font Awesome ren
 
 > [!NOTE]
 > Font Awesome 7 also gives every icon the same width by default. Homer keeps the previous behaviour, so nothing moves after upgrading. See [icon width](configuration.md#icon-width) if you want the new look.
+
+## My custom stylesheet stopped styling the service cards
```
bastienwirtz:
```suggestion
## My custom stylesheet stopped styling the service cards since 26.09.20
```
(exact version to be updated before merge & release)

docs/troubleshooting.md
```diff
@@ -68,3 +68,20 @@ Only one icon is lost: **`fa-vector-square` no longer exists**. Font Awesome ren
...
+- `.tag` collapses with `max-width` rather than `width`, so a rule setting `width` has no effect.
+- `.quicklinks` is a flex lane at the bottom of the card, so `float` no longer applies.
+- `.card-lane` reaches both card edges and clips what overflows them, so anything positioned inside it stops at the card. The tag sits at the lane's inline end rather than overhanging by its own offset.
+- The status dot animation is named `status-pulse`.
+
+`.card:hover` still matches, so hover rules keyed on the card itself keep working.
+
+## I cannot select card text, and some tooltips stopped appearing
+
+The whole card is one link, drawn as a transparent overlay on top of the content. That overlay takes the pointer, so text inside a card can no longer be selected, and a native `title` tooltip on an element in the card body no longer opens. Tooltips on the badges and the tag are unaffected, as those sit above the overlay.
```
bastienwirtz, line 87: > The whole card is one link, drawn as a transparent overlay on top of the content.

I'm not super fan of that, but i think I see why it's necessary / easier that way, it should be all right.

bastienwirtz (approved):

docs/troubleshooting.md
tanasegabriel: Done!

docs/troubleshooting.md
tanasegabriel: Honestly I went back and forth on this one too, so I'm kinda glad it lands as reasonable.

The main concern here was that an `<a>` wrapped around the card content can't hold the quicklinks and trying to address that opens a can of worms / ends up in a game of "whack a mole" with the layout. Doing it this way felt like the better fix, but open to discuss it.

PR #1081, opened 2026-09-04
feat(cards): move every card onto the shared API

### LINKED ISSUE #953: Make WUD badge green when no updates are available
  
Right now, the WUD extension always shows an orange badge with the number of running containers, even when there are no updates. This looks like a warning even when everything is OK. 
 
[See related code](https://github.com/bastienwirtz/homer/blob/c230392da8a30a7a6995bea0e9180e261fedd635/src/components/services/WUD.vue#L79C5-L79C15)

– Show a green badge with the number of running containers if there are no updates.  
– If updates exist, show a badge for update count in orange or red (based on number).  
– Keep `?` for WUD API errors.

– Leaving it as-is, but orange always feels like a warning.  
– Showing both counts all the time, but that makes the UI too busy.

**Additional context**

Current behavior (no updates, still orange):  

Image

**Happy to make a PR for this if needed.**

tanasegabriel:
Second of two stacked PRs. The first gave [`Generic.vue`](https://github.com/bastienwirtz/homer/blob/daa017d/src/components/services/Generic.vue) the card layout and kept the old slots open so nothing had to change at once. This one migrates every card onto the new props, moves the shared loading and error handling into the [`service` mixin](https://github.com/bastienwirtz/homer/blob/daa017d/src/mixins/service.js), and removes the compatibility the first PR held open.

It deletes about 2000 lines more than it adds.

## What changes

- **Cards feed props instead of positioning markup.** `<Generic :item="item" :subtitle="subtitle" :status="status" :badges="badges" />` replaces the `#content` and `#indicator` slots. Richer content still goes through the `icon` / `subtitle` / `aside` / `badges` slots.
- **One vocabulary for "is it up".** `status` is `{ state, label }` with `state` one of `online`, `offline`, `warning`, `busy` or `unknown`, replacing the six sets of words cards invented for themselves, and every colour now comes from a theme token.
- **Counters become badges.** `badges` is a list of `{ key, label, value, tone }` with `tone` one of `info`, `success`, `warning`, `danger`, `accent` or `neutral`. Empty values, zeros without `showZero` and any `key` listed in the item's `hide` are dropped centrally, and counts cap at `99+`.
- **The mixin owns fetching and failure.** It auto-wires `fetchData()` on create and on the `updateIntervalMs` schedule, and adds `load()`, `serverError`, `reachabilityStatus()`, `connectionBadge()`, `requireConfig()` and `versionSubtitle`. A card that cannot reach its API now says so once, the same way everywhere.
- **Formatting leaves the mixin and the cards.** `src/utils/format.js`, introduced in the first PR for `capCount`, gains `displayRate`, `displaySize` and `displayDuration`, the last two lifted out of `Immich` and `OctoPrint`. They are pure functions with no `this`, so the mixin keeps only what needs component state and `Generic` imports what it needs without the mixin. [`Gotify`](https://github.com/bastienwirtz/homer/blob/daa017d/src/components/services/Gotify.vue)'s message count caps at `99+` with everything else, rather than at its own `100`.
- **An ESLint rule enforces the contract**, so the next card starts on it rather than inventing a seventh vocabulary: a single `<Generic>` root, known slots, known `tone` and `state` values, no redeclared `item` prop, a declared `serverError`, and no card restyling the shared chrome.
- [`docs/development.md`](https://github.com/bastienwirtz/homer/blob/daa017d/docs/development.md), [`docs/customservices.md`](https://github.com/bastienwirtz/homer/blob/daa017d/docs/customservices.md) and `AGENTS.md` document the props, the slots and the rule.

Fixes #953.

## What it removes

Everything the first PR held open, all of it marked `TODO(cards-migrate-to-api)` there:

- the `#content` and `#indicator` slots
- the pre-migration status names `ready`, `error`, `pending` and `in-progress`, and `--status-color: transparent`
- the dot spacing as a `margin-inline-end`, back to a `gap` on `.status` now that no card ships a margin of its own
- `z-index` on `.card-aside`, so **the status chip is clickable again**. It was held above the card link so unmigrated indicators kept the `title` tooltips that were the only labels on their counts. Badges carry their own tooltips without blocking the click.
- the placeholder in `src/utils/format.js`, now that the remaining formatters have joined `capCount` there

Along with the per-card styling that duplicated the shared chrome: 21 scoped `.status` blocks, 25 scoped `.notifs` blocks, and all but two of the hardcoded hex colours. The two that stay are [`OpenWeather`](https://github.com/bastienwirtz/homer/blob/daa017d/src/components/services/OpenWeather.vue)'s icon backdrop, which is a backdrop rather than a theme colour.

## Unreachable APIs now report themselves

Running each card against a real instance rather than fixtures showed that a handful could look healthy, or say nothing at all, while their API was unreachable. That is what happens when every card decides for itself what a failed fetch means.

Routing them through `load()`, which records the failure, and `reachabilityStatus()` or `connectionBadge()`, which surface it, fixes those cases without patching any of them individually, and covers the next card by default.

## Compatibility

- **A configured [`subtitle`](https://github.com/bastienwirtz/homer/blob/daa017d/docs/customservices.md) is now honoured everywhere.** Six cards rendered their own subtitle without checking `item.subtitle` first, and [`Gatus`](https://github.com/bastienwirtz/homer/blob/daa017d/src/components/services/Gatus.vue) rendered both at once. On those seven a `subtitle:` that used to do nothing, or to appear alongside the card's own line, now replaces it, which is [the documented behaviour](https://github.com/bastienwirtz/homer/blob/daa017d/docs/customservices.md#L90).
- **Status and badge colours come from the theme.** A card that shipped its own green now uses `--status-online`, so a custom stylesheet targeting a card's scoped `.status` or `.notif` rules no longer matches. The tokens are listed in [theming](https://github.com/bastienwirtz/homer/blob/daa017d/docs/theming.md).
- **No card styles the shared chrome any more.** Anyone maintaining an out-of-tree card that copied a scoped `.status` or `.notifs` block loses their reference implementation; the props and the skeleton in `docs/development.md` replace it.

## Please merge both PRs in the same release

The first one alone looks half-migrated: the new chrome with cards still using their own status words, colours and pill placement. The split exists so the layout can be reviewed on its own, not so the two can ship separately.

## Testing

- `pnpm lint` and `pnpm build` clean, with the new rule active.
- Every card exercised against a real service in a container, with authentication configured wherever it could be, rather than against fixtures alone. That is what surfaced the unreachable-API cases above.
- Rendered against a config covering every card plus the chrome cases, side by side with `main` and with the first PR alone, in light and dark, in both layouts, across the classic, walkxcode and neon themes.
- Confirmed no card renders its own subtitle markup, none overrides `#content`, no scoped `.status` or `.notifs` block remains, and the status dot colours resolve to the same values `main` used, so no card's dot changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,15 +22,16 @@ Homer is a static Vue.js 3 PWA dashboard that loads configuration from YAML file
 - **Entry Point**: `src/main.js` mounts the Vue app
 - **Root Component**: `src/App.vue` handles layout, configuration loading, and routing
 - **Configuration System**: YAML-based with runtime merging of defaults (`src/assets/defaults.yml`) and user config (`/assets/config.yml`)
-- **Service Components**: 53 specialized integrations in `src/components/services/` that extend a Generic component pattern
+- **Service Components**: 58 specialized integrations in `src/components/services/` that extend a Generic component pattern
 
 ### Service Integration Pattern
 
 All service components follow this architecture:
 
-- Extend `Generic.vue` using Vue slots (`<template #indicator>`, `<template #content>`, `<template #icon>`)
-- Use the `service.js` mixin (`src/mixins/service.js`) for common API functionality
+- Render a single `Generic.vue` as the template root and feed it the `subtitle`, `status` and `badges` props, using a slot (`icon`, `subtitle`, `aside`, `badges`) only for richer content. `Generic` owns the card layout and `pnpm lint` enforces it.
+- Use the `service.js` mixin (`src/mixins/service.js`) for common API functionality. It supplies the `item` prop, calls your `fetchData()` on create and on the refresh schedule, and owns error state through `load()` and `serverError`.
 - Use a custom `fetch` method provided by the service mixin to seamlessly support proxy configuration, custom headers, and credentials.
+- Keep display formatting out of the mixin. `capCount`, `displayRate`, `displaySize` and `displayDuration` are pure functions in `src/utils/format.js`, importable by anything, including components that are not cards. Anything needing `this` (the item config, the proxy, the lifecycle) belongs on the mixin instead.
 - Never call the native `fetch` in a service component unless you have a specific reason not to. See `OpenWeather` (third-party public API) or `Rtorrent` (XML-RPC on a separate host) for use case where the service fetch is not suitable.
 - Pass the service's own headers (an API key, an `Authorization` header, ...) as the `init.headers` argument of `this.fetch(path, init, json)`. They layer over the user's `proxy.headers` and item `headers` rather than replacing them, so a card must never build its headers from `proxy` or `item.headers` itself. Names are compared case-insensitively, and the card value wins on conflict.
 
@@ -40,12 +41,14 @@ Services support automatic data refreshing using a centralized scheduler system 
 
 #### Global Configuration
 
-`autoUpdate: 30000` - Set default interval for all services (30 seconds)
+`updateIntervalMs: 30000` - Set default interval for all services (30 seconds)
 
 #### Service Configuration
-- **Service-specific interval**: `updateInterval: 10000` - Override global default
-- **Disable per service**: `autoUpdateInterval: false` - Disable for specific service
-- **Use global default**: Omit `autoUpdateInterval` to use global setting
+- **Service-specific interval**: `updateIntervalMs: 10000` - Override global default
+- **Disable per service**: `updateIntervalMs: false` (or `0`) - Disable for specific service
+- **Use global default**: Omit `updateIntervalMs` to use global setting
+
+The mixin still accepts `checkInterval`, `downloadInterval`, `rateInterval`, `torrentInterval` and `updateInterval`, logging a deprecation warning for each.
 
 ### Configuration & Routing
 
--- a/docs/customservices.md
+++ b/docs/customservices.md
@@ -85,9 +85,54 @@ Available services are located in `src/components/`:
   endpoint: https://my-service-api.url # Optional: alternative base URL used to fetch service data when necessary.
   useCredentials: false # Optional: Override global proxy.useCredentials configuration.
   headers: # Optional: Per-item headers, layered over (and overriding) proxy.headers. A card setting the same header wins over both.
+  hide: [] # Optional: List of keys to remove from this card, badges and card-specific fields alike (see each card below for the keys it offers).
 ```
 
 If a subtitle is provided, (using the `subtitle` configuration key), **it will override (hide)** any custom information displayed on the subtitle line by the custom integration.
+
+Every smart card uses the same layout: icon, name and subtitle, a **status chip** on the right, **badges** along the top, and the `quick` links and `tag` sharing the bottom. A card only shows the parts it has data for.
+
+- **Status chip**: green is healthy, red down, amber needs attention, pulsing green busy, grey unknown. Nothing shows until the first request settles.
+- **Badges**: blue activity, green healthy totals, amber warnings, red errors, purple a secondary count, grey a plain total. Zero is hidden, counts above 99 read `99+`, and `?` means the API was unreachable. Hover one for its meaning.
+
+<details>
+<summary>Keys for the <code>hide</code> option</summary>
+
+Mostly badge keys; a few cards route their own fields through `hide` too.
+
+| card | keys |
+| --- | --- |
+| AdGuard Home, Docuseal, Emby / Jellyfin, Gatus, Gitea, Gotify, Home Assistant, Matrix, Nextcloud, OctoPrint, OliveTin, OpenHAB, PeaNUT, Pi-hole, Ping, Prometheus, Traefik, TrueNAS Scale, Uptime Kuma, Vaultwarden, Wallabag | no badges, these cards show a status chip |
+| Docker Socket Proxy | `running`, `stopped`, `errors`, `serverError` |
+| FreshRSS | `subscriptions`, `unread`, `serverError` |
+| Glances | `serverError` |
+| Healthchecks | `up`, `down`, `grace`, `serverError` |
+| HyperHDR | `running`, `stopped` |
+| Immich | `users`, `photos`, `videos`, `usage`, `serverError` |
+| Jellystat | `streams`, `serverError` |
+| LibrisLog | `reading`, `want-to-read`, `serverError` |
+| Lidarr | `activity`, `missing`, `warnings`, `errors`, `serverError` |
+| Mealie | `serverError` |
+| Medusa | `news`, `warnings`, `errors`, `serverError` |
+| Miniflux | `unread`, `serverError` |
+| Mylar | `wanted`, `upcoming`, `serverError` |
+| NetAlertx, PiAlert | `total`, `connected`, `newdevices`, `downalert`, `serverError` |
+| Paperless-ng | `serverError` |
+| Plex | `streams`, `series`, `movies`, `warnings`, `errors`, `serverError` |
+| Portainer | `running`, `dead`, `misc` |
+| Prowlarr | `warnings`, `errors`, `serverError` |
+| Proxmox | the subtitle fields `vms`, `vms_total`, `lxcs`, `lxcs_total`, `disk`, `mem`, `cpu`, plus `serverError` |
+| Radarr, Readarr, Sonarr | `activity`, `missing`, `warnings`, `errors`, `serverError` |
+| SABnzbd | `downloads`, `serverError` |
+| Scrutiny | `passed`, `failed`, `unknown`, `serverError` |
+| Seerr | `media`, `pending`, `processing`, `issues`, `serverError`, plus the subtitle icons `updateAvailable` and `restartRequired` |
+| Speedtest Tracker | `serverError` |
+| Tautulli | `streams`, `serverError` |
+| Tdarr | `queue`, `errored`, `serverError` |
+| Transmission, qBittorrent, rTorrent | `torrents`, `serverError` |
+| WUD | `running`, `update`, `serverError` |
+
+</details>
 
 > [!TIP]
 > **Auto refresh of the card data**: Some cards support periodic update (see indication in detail below). It can be enabled or disabled globally for all service, or individually for each service using the `updateIntervalMs` configuration option.
--- a/docs/development.md
+++ b/docs/development.md
@@ -12,11 +12,22 @@ pnpm dev
 Custom services are small VueJs component (see `src/components/services/`) that add little features to a classic, "static", dashboard item. It should be very simple.
 A dashboard can contain a lot of items, so performance is very important. 
 
-The [`Generic`](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Generic.vue) service provides a typical card layout which
-you can extend to add specific features. Unless you want a completely different design, extended the generic service is the recommended way. It gives you 3 [slots](https://vuejs.org/v2/guide/components-slots.html#Named-Slots) to extend: `icon`, `content` and `indicator`. 
-Each one is **optional**, and will display the usual information if omitted. Content given to `icon` must be a `.card-icon` element: the card is a grid, and anything else lands outside the icon zone.
+Each service must bind the `item` [property](https://vuejs.org/v2/guide/components-props.html) to the [`Generic`](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Generic.vue) component, which owns the card layout.
 
-Each service must implement the `item` [property](https://vuejs.org/v2/guide/components-props.html) and bind it the Generic component if used.
+```Vue
+<Generic :item="item" :subtitle="rttLabel" :status="reachabilityStatus()" :badges="badges" />
+```
+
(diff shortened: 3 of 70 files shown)


docs/customservices.md
```diff
@@ -85,10 +85,55 @@ Available services are located in `src/components/`:
...
 
 If a subtitle is provided, (using the `subtitle` configuration key), **it will override (hide)** any custom information displayed on the subtitle line by the custom integration.
 
+Every smart card uses the same layout: icon, name and subtitle, a **status chip** on the right, **badges** along the top, and the `quick` links and `tag` sharing the bottom. A card only shows the parts it has data for.
+
+- **Status chip**: green is healthy, red down, amber needs attention, pulsing green busy, grey unknown. Nothing shows until the first request settles.
+- **Badges**: blue activity, green healthy totals, amber warnings, red errors, purple a secondary count, grey a plain total. Zero is hidden, counts above 99 read `99+`, and `?` means the API was unreachable. Hover one for its meaning.
+
+<details>
+<summary>Keys for the <code>hide</code> option</summary>
```
bastienwirtz: Imo, we should document the hide keys for each services, since even if a significant portion of services has the same keys, it's natural that different service expose different kind of information. It should be service specific. 
Plus this adds a section to maintain, and on the other hand when you look at a specific service below, you don't have all the information.

docs/development.md
```diff
@@ -12,11 +12,22 @@ pnpm dev
 Custom services are small VueJs component (see `src/components/services/`) that add little features to a classic, "static", dashboard item. It should be very simple.
 A dashboard can contain a lot of items, so performance is very important. 
 
-The [`Generic`](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Generic.vue) service provides a typical card layout which
-you can extend to add specific features. Unless you want a completely different design, extended the generic service is the recommended way. It gives you 3 [slots](https://vuejs.org/v2/guide/components-slots.html#Named-Slots) to extend: `icon`, `content` and `indicator`. 
-Each one is **optional**, and will display the usual information if omitted. Content given to `icon` must be a `.card-icon` element: the card is a grid, and anything else lands outside the icon zone.
+Each service must bind the `item` [property](https://vuejs.org/v2/guide/components-props.html) to the [`Generic`](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Generic.vue) component, which owns the card layout.
```
bastienwirtz: I think encouraging the standard layout is great, however, I wouldn't say it's necessary. the slots still exists and provides a way to get more freedom when it could be useful. 
Maybe something like: 
```suggestion
The [`Generic`](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Generic.vue) service provides a standard card layout. Each service should bind the `item` [property](https://vuejs.org/v2/guide/components-props.html) to the [`Generic`](https://github.com/bastienwirtz/homer/blob/main/src/components/services/Generic.vue) component, to display custom information.

Unless you need a different  structure, extend the generic service and binding the properties is the recommended way. Otherwise It possible to extend the available [slots](https://vuejs.org/v2/guide/components-slots.html#Named-Slots) of the Generic card, or even a do a entirely custom structure. Keep in mind that **the goal is to keep a consistent design between all cards** so everything looks good together. 
```

What do you think ?

bastienwirtz: For exemple, I might want to add a drop down button to enable / disable protection in piHole, or add a minimalist css progress bar for whatever reason. It those case, you need to inject custom html in the slots.

docs/development.md
```diff
@@ -12,11 +12,22 @@ pnpm dev
...
+- **`subtitle`** (string): the subtitle line. A config `subtitle` wins over it.
+- **`status`** (`{ state, label }`): the chip on the right. `state` is `online`, `offline`, `warning`, `busy` or `unknown`; `label` is optional.
+- **`badges`** (`[{ key, label, value, tone }]`): the counters along the top. `tone` is `info`, `success`, `warning`, `danger`, `accent` or `neutral`; `label` is the tooltip. Empty values, zeros (unless `showZero: true`) and any `key` in the item's `hide` are dropped, and counts cap at `99+`.
+
+The `icon`, `subtitle`, `aside` and `badges` slots take richer content, each replacing its zone.
+
+A card must surface an API failure exactly once: through its status chip if it has one, otherwise with a `connectionBadge()`.
+
+> [!IMPORTANT]
+> `pnpm lint` enforces this: a single `<Generic>` root, known slots, known `tone` / `state` values, no redeclared `item` prop, no `created()` that only wires `autoUpdateMethod`, a declared `serverError`, and no card restyling the shared chrome. Name a lookup table `_STATE` or `_TONE` in SCREAMING_CASE to have its values checked too.
```
bastienwirtz: Not sure if we should enforce it (see my previous comment)

AGENTS.md
```diff
@@ -22,15 +22,16 @@ Homer is a static Vue.js 3 PWA dashboard that loads configuration from YAML file
 - **Entry Point**: `src/main.js` mounts the Vue app
 - **Root Component**: `src/App.vue` handles layout, configuration loading, and routing
 - **Configuration System**: YAML-based with runtime merging of defaults (`src/assets/defaults.yml`) and user config (`/assets/config.yml`)
-- **Service Components**: 53 specialized integrations in `src/components/services/` that extend a Generic component pattern
+- **Service Components**: 58 specialized integrations in `src/components/services/` that extend a Generic component pattern
```
bastienwirtz: This changes continuously, we should remove the number

src/mixins/service.js
```diff
@@ -39,6 +42,13 @@ export default {
     if (this.endpoint && this.endpoint.endsWith("/")) {
       this.endpoint = this.endpoint.slice(0, -1);
     }
+
+    // A card's own created() runs after this one, so it can still override
+    // autoUpdateMethod to pick a method at runtime.
+    if (typeof this.fetchData === "function") {
+      this.autoUpdateMethod = this.fetchData;
+      this.fetchData();
+    }
```
bastienwirtz, line 45: I think this induce a bit too much of magic. 
The behavior depends on the function name, I'm not fan of it.
The way custom data if fetched is the core a of a custom card, and it should be "visible". 
Also, and that's a important point imo, all information are not worth auto updating. Auto update runs parallel requests in the background, and the less there is, the better. A dynamic value should be auto updated, a static one (or not static but not changing frequently) like version should not be updated.

src/mixins/service.js
```diff
@@ -30,6 +30,9 @@ export default {
     globalConfig() {
       return this.config() || {};
     },
+    versionSubtitle() {
+      return this.versionstring ? `Version ${this.versionstring}` : null;
+    },
```
bastienwirtz: This is hidden logic, it should remains in each cards implementation.  
The mixin is a low level helper and should not involve the logic related to the content.  
```suggestion
```

src/mixins/service.js
```diff
@@ -52,7 +62,45 @@ export default {
...
+      let failed = false;
+      await Promise.all(
+        requests.map((request) =>
+          request.catch((e) => {
+            console.error(e);
+            failed = true;
+          }),
+        ),
+      );
+      this.serverError = failed;
```
bastienwirtz, line 77: I don't think this adds a lot of value, while it can limit the way cards can handle errors in there own way.

AGENTS.md
tanasegabriel: That's a good shout; done!

src/mixins/service.js
tanasegabriel: Good spot, agreed.

i still think this might be useful, so I'm proposing using `utils/`. `versionSubtitle` is gone from the mixin. `displayVersion(version)` now sits in `src/utils/format.js` next to `capCount` and the other `display*` helpers, as a pure function with no `this`.

Cards use it the way `OctoPrint` already uses `displayDuration`: import it, expose it through `methods`, call it from the template.

```vue
<Generic :item="item" :subtitle="displayVersion(versionstring)" :status="reachabilityStatus()" />
```

That keeps the mixin clear of anything about what a card displays, while still leaving one spelling of the version line instead of eleven.

src/mixins/service.js
tanasegabriel: Ok, I've changed the approach.

On `main`, 40 cards set `autoUpdateMethod` and 14 deliberately do not: they call their loader once in `created()` and never refresh. Auto-wiring `fetchData` switched polling on for all 14, eight of which fetch nothing but a version string. It also quietly made the "Auto refresh is supported by this integration" note meaningless, since it was true of 34 of 53 sections and would now be true of all of them.

So the mixin no longer treats the two as the same decision. It calls `fetchData()` once on create, and refreshing is separate:

```js
initAutoUpdate() {
  if (this.$options.autoUpdate === false) return;
  const method = this.autoUpdateMethod ?? this.fetchData;
  ...
}
```

`Docuseal`, `Gitea`, `Matrix`, `OliveTin`, `Traefik`, `TrueNAS Scale`, `Vaultwarden` and `Wallabag` declare `autoUpdate: false` and are never polled. Everything else refreshes, which now includes seven cards that did not on `main` but arguably should have, `SpeedtestTracker` most obviously. `PiHole` and `Seerr` keep their `created()` for the runtime method choice. The docs state the rule once and name the eight exceptions, so the claim is checkable.

The one part I'd gently push back on is the name-based bit, `fetchData` getting picked up because of what it's called. I can see why that reads as magic. But the fetch itself is still right there in the card; what moved into the mixin is only the two lines that register it, and `fetchData` felt like the same kind of convention as `data` or `created`. Spelling it out again means a `created()` in all 53 cards that have one, so roughly 200 lines that all say the same thing. If you disagree with all of that, let me know. I just don't think it buys much now that the polling is decoupled from it.

docs/customservices.md
tanasegabriel: Yeah, fair point, moved them. Each service's section now carries its own `hide` line in the YAML example:

```yaml
  hide: [] # Optional: badges to hide: "up", "down", "grace", "serverError".
```

The big table is gone. What's left in Common options is only the genuinely shared bit: that `hide` takes badge keys, and that every card hitting an API also offers `serverError`.

One thing that popped up while moving them: `Mylar` had no section in this file at all, so that table was the only place its keys were ever documented. I've added a short section for it rather than lose them.

Worth flagging a related change in the same file, since it goes the other way and you might not agree. The per-card "Auto refresh is supported by this integration" line is gone from all 34 sections that had it, replaced by one statement at the top plus the list of exceptions.

It went the opposite way to `hide` because the answer stopped being per-card. `hide` keys genuinely differ per service, which is your point and why they moved down. Refresh doesn't any more: after this PR every card refreshes except the eight that only read a version string. The old note was true nearly everywhere, and it was also missing from seven sections where it had just become true (`SpeedtestTracker`, `Nextcloud`, `Mealie`, `OpenHAB`, `HyperHDR`, `PaperlessNG`, `HomeAssistant`), so it needed touching either way. Happy to put all 34 back and add the missing seven instead if you'd rather keep it uniform.

docs/development.md
tanasegabriel: Yeah, "must" might be too strong, so I've taken your wording almost as-is.

I've also added a mechanism that might prove useful here: a way for a card to opt out on purpose. It's an in-file marker naming which checks it's dropping, and why:

```vue
<!-- card-contract: off structure -- the protection toggle needs its own markup -->
```

Groups are `structure`, `vocabulary`, `mixin`, `errors` and `style`, so PiHole could take a completely custom structure while still being held to the shared `tone` / `state` values.

On the dropdown itself, I'd actually use it as an argument *for* the shared layout rather than against it. A protection toggle isn't really PiHole-specific; Uptime Kuma for example has the same shape of problem with maintenance mode. If the first one lands as a card-local one-off, the second reinvents it. If it lands as a control the shared card knows about, the second one is just a prop. That's the reuse I think would be beneficial., rather than uniformity for its own sake.

docs/development.md
tanasegabriel: Fair - the docs now present it as the default with a named way out, and the `[!IMPORTANT]` callout is gone. But I would still suggest we make use of this?

My reasoning for keeping it: there are no tests for this in the repo, so the rule is the only thing that notices when a card drifts from the shared layout, and that drift is what produced the bugs this PR fixes (e.g. quicklinks unreachable on any card that overrode `#content`, some cards showing healthy with a dead API). A new card complies for free, and an outlier is one commented line away.

Let me know your thoughts please!

src/mixins/service.js
tanasegabriel: I'd push back a little here, though with an escape hatch attached.

They're opt-in helpers rather than enforcement: nothing stops a card catching its own way, and eleven currently do. What running each card against a real instance showed is that when every card decides for itself, some of them decide wrong. A few looked healthy, or said nothing at all, while their API was unreachable. `load()` and `reachabilityStatus()` are there so the next card gets that right by default, not to take the option away.

The only genuinely enforced part is the lint check that a card surfaces failure somehow, and that now has a documented exit (`card-contract: off errors -- <reason>`).

For what it's worth this came from [your suggestion in #1065](https://github.com/bastienwirtz/homer/pull/1065#discussion_r3652695035), when Seerr grew its own `cap()`. And with no tests here, that consistency is doing real work: before this PR the same idea was spelled three different ways across the cards (console only, a subtitle error string, an error badge) - if you feel strongly about this, let me know please.

PR #1086, opened 2026-09-09
fix: prevent browsers from serving a stale configuration

### LINKED ISSUE #1084: Front-end not reflecting changes I made anymore?
**Describe the bug**
If I `curl -I <url>`, I see that the "last-modified" date says `Sun, 30 Aug 2026 16:19:04 GMT` even though I've just modified something inside the yaml file (and it's Mon, 07 Sept 2026) but I'm not even sure that's the root cause of the problem.

My problem is mainly the fact that I changed something inside the yaml files and it was not reflected on the website/front-end even with refreshing and everything. It showed fine in a private browser window but not if I kept refreshing my regular browser.

**Expected behavior**
I expected my changes to be visible on the front-page/website.

I am not sure if this is the root cause of the issue but the last-modified date not updating is something I've realized now and suspect it might lead to the browser not loading new data but instead showing the cached data.

**Logs & errors**
I do not believe this is applicable

**Screenshots**
I do not believe this is applicable

**Configuration**
I just changed the URL of a service I have in my config like this:
```yml
services:
  - name: "My Services"
    icon: "fas fa-vote-yea"
    items:
      - name: "Some Name"
        icon: "fas fa-hand-holding-usd"
        subtitle: "Some subtitle"
        url: "https://foo.bar" # This is the URL which I had changed but then it wasn't reflected on the website itself
```

tanasegabriel:
## Description

Homer can sometimes serve stale `config.yml` because neither the app nor the bundled lighttpd told browsers how long to reuse it, so they fell back to heuristic freshness and could serve an edited file from cache for hours without revalidating. The same applies to `index.html` and everything else under `assets/`.

This adds `cache: "no-cache"` to the config fetch in `App.vue` so the browser always revalidates, and states the caching policy explicitly in `lighttpd.conf`: content-hashed files under `resources/` are immutable, everything else is revalidated on every load. Unchanged files still cost only a `304`.

Fixes https://github.com/bastienwirtz/homer/issues/1084

## Type of change

- [ ] Refactoring

## Checklist:

- [ ] I have made corresponding changes to the documentation (`README.md`).


lighttpd.conf
```diff
@@ -10,7 +10,16 @@ server.indexfiles      = ("index.html")
 server.follow-symlink  = "enable"
 server.feature-flags  += ( "server.clock-jump-restart" => 0 )
 server.max-request-field-size = 65535
-accesslog.filename = "/dev/fd/3" 
+accesslog.filename     = "/dev/fd/3"
+etag.use-inode         = "disable"
```
tanasegabriel, line 14: `lighttpd` includes the file's inode in its etag by default, so byte-identical files get different etags on different machines. That didn't matter while nothing revalidated, but now that everything under `assets/` revalidates on every load, a per-replica etag would turn those cheap `304`s back into full `200` responses for anyone running more than one Homer replica behind a load balancer.

nginx has never seem to have included the inode in its etag, and Apache dropped it from the default in 2.4, apparently for the same reason.

